Issue #45, opened 2026-05-07
User Conversation History


Perimora:
## User story
1. As a user
2. I want my converstations to be stored persistantly 
3. So that I can revisted them later

## Acceptance criteria
* A conversation is automatically created when the user sends the first message
* All user and system messages are stored in the backend
* Reloading the application restores previous conversations
* Conversations are uniquely id (e.g. conversation_id)
* If user data was extracted during the conversation it is correctly displayed in the conversation history via FE. To ensure transparency, it must be clear to the user which data was extracted by which tool.

## Definition of done (DoD)
* General coding standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

Issue #18, opened 2026-04-21
Basic LLM Integration 


Caerum5:
## User story
As an end user
I want my transcribed input to be processed by an LLM
So that I receive a meaningful response from the conversational agent

## Acceptance criteria
* The backend integrates an LLM client configurable for either Gemini or OpenAI API
* The LLM client accepts a single text prompt as input
* The text prompt may come from the Speech-to-Text component (#14) or from a mocked transcript until #14 is completed
* The generated text response from the LLM is successfully received by the backend
* The prompt instructs the LLM to respond in the same language as the input
* The LLM response is made available to the Text-to-Speech component (#15)
* The response may be passed to a mocked Text-to-Speech interface until #15 is completed
* Basic error handling is implemented for API failure, timeout, or missing configuration
* The implementation is aligned with the documented system architecture (System Architecture #12)
* The LLM integration is documented in the project wiki, including required .env variables

## Definition of done (DoD)
* Generell Coding Standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written.

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

PR #46, opened 2026-05-07
feat: DB Integration - User and ChatHistory models, CRUD operations and migrations

### LINKED ISSUE #32: DB Integration
## User story
1. As a software developer
2. I need to setup a Database
3. So that basic workflows can be stored persistently 

## Acceptance criteria
* A database system is selected (e.g. PostgreSQL) and configured for local development (e.g. via Docker compose)
* A minimal schema is implemented for:
  * users (id, identifier)
  * chat_history (id, user_id, message, timestamp, role)
* The backend can connect to the database and perform basic operations (create and read).
* Basic migration handling is implemented (e.g. using a migration tool or simple schema initialization).
* Basic error handling is implemented for database connection failures.
* The database setup and schema are documented in the project wiki and credentials/connection strings are managed in .env
* The implementation is aligned with System Architecture #12

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

1Endri:
Closes #32

## What was done
- Added User model (id, username, email, hashed_password, is_active, created_at, updated_at)
- Added ChatHistory model (id, user_id, conversation_id, role, message, timestamp)
- Added relationship between Conversation and ChatHistory
- Created CRUD operations for User and ChatHistory with error handling and rollback
- Generated and applied Alembic migration for new tables
- Written 10 unit tests (7 User CRUD, 3 ChatHistory CRUD) - all passing

## How to test
docker compose up -d postgres
cd backend
poetry run alembic upgrade head
poetry run pytest tests/test_crud.py -v


backend/migrations/versions/6765689d3964_add_user_and_chat_history_tables.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+# revision identifiers, used by Alembic.
+revision: str = '6765689d3964'
+down_revision: Union[str, Sequence[str], None] = 'd9d0cb2dc240'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
```
Asadraza17: Looks like there are two migration files for the same change - d9d0cb2dc240_ has the actual op.create_table calls for users + chat_histories, and this one (6765689d3964_) is an empty pass-body artifact that just depends on the first. Should be deleted before the merge so the migration chain stays clean.

backend/tests/test_crud.py
```diff
@@ -0,0 +1,129 @@
...
+from taskorbit.database.crud import (
+    create_user, get_user, get_user_by_email, get_user_by_username,
+    get_users, update_user, delete_user,
+    create_chat_history, get_chat_histories_by_user,
+    get_chat_histories_by_conversation, delete_chat_history
+)
+
+
+# Use in-memory SQLite for testing
+TEST_DATABASE_URL = "sqlite:///./test.db"
```
Asadraza17: Worth switching to sqlite:///:memory: so the test DB doesn't leave a test.db file in the repo root if a test crashes mid-run. Right now, Base.metadata.drop_all cleans the tables between tests, but the file itself persists, and a crashed test can leave a stale schema that confuses the next run.

backend/src/taskorbit/database/crud.py
```diff
@@ -0,0 +1,217 @@
...
+from .models import User, ChatHistory, Conversation
+
+
+# ============ USER CRUD ============
+
+def get_user(db: Session, user_id: int) -> User | None:
+    """Get a user by ID."""
+    try:
+        return db.query(User).filter(User.id == user_id).first()
+    except SQLAlchemyError:
```
Asadraza17: The DoD for #32 lists "Logging" as a coding standard, and the rest of the backend (livekit.py, orchestration/init.py) uses structlog for everything that fails. Right now we swallow every SQLAlchemyError silently and return None.. if something goes wrong in production we have no trail of what happened. Worth adding a logger.error("user_lookup_failed", ...) call before the return at minimum, so downstream tickets aren't debugging blind

backend/src/taskorbit/database/models.py
```diff
@@ -64,3 +67,43 @@ class ToolExecution(Base):
...
+    updated_at: Mapped[datetime | None] = mapped_column(
+        DateTime(timezone=True), onupdate=func.now()
+    )
+
+    chat_histories: Mapped[list[ChatHistory]] = relationship(
+        back_populates="user", cascade="all, delete-orphan"
+    )
+
+
+class ChatHistory(Base):
```
Asadraza17, line 92: Quick architectural question: chat_histories here and conversation_messages already on main both store role + message content + conversation_id + timestamp. The difference is that chat_histories has a user_id link. As we wire downstream tickets like #18, #45, #47, which one is the canonical message store? If both stay, we need a clear rule on what goes where.

backend/src/taskorbit/database/crud.py
```diff
@@ -0,0 +1,217 @@
+"""CRUD operations for database models."""
+
+from sqlalchemy.orm import Session
```
Asadraza17, line 3: Quick heads-up: this is not blocking. The session is synchronous, but the rest of the backend (FastAPI routes, process_message in orchestration) is asynchronous. When downstream tickets like #18, #45, #47 wire this CRUD into async routes, the sync DB ops will block the event loop. Worth a call here on whether to migrate to AsyncSession now or absorb the cost later.

Asadraza17 (commented):
Solid foundation overall - five of the seven ACs cleanly pass, the migrations work, ten unit tests with rollback handling in CRUD, and the Database Setup wiki page is in place. Schema additions for users and chat_histories cover the AC, and you linked chat_histories to the existing conversations table, which is good forward thinking.

Two things worth fixing before merge - the duplicate empty migration file (looks like an autogen artifact) and the test.db file leak in the test fixture (easy switch to in-memory SQLite).

Two architectural questions worth a team alignment - the silent error swallowing (we lose visibility into DB failures) and the overlap between chat_histories and conversation_messages. Plus, a quick DoD note: "Logging" is one of the coding standards listed on the DoD, and crud.py currently has none of it. Worth at least a logger.error before each silent return, so we have a trail when something fails.

backend/migrations/versions/6765689d3964_add_user_and_chat_history_tables.py
1Endri: done

backend/tests/test_crud.py
1Endri: done

backend/src/taskorbit/database/crud.py
1Endri: done

backend/src/taskorbit/database/models.py
1Endri: done

backend/src/taskorbit/database/crud.py
1Endri: done

Asadraza17 (approved):
Approving as all code asks addressed. Flagging for team sync: ChatHistory vs ConversationMessage overlap is still unresolved in models.py. Worth a quick alignment before #18/#45/#47 pick this up so we don't end up with messages split across both tables

Asadraza17 (approved):

PR #58, opened 2026-05-10
Feature/33 basic agent runtime

### LINKED ISSUE #33: Basic Agent Runtime
## User story
1. As a Developer
2. I want to implement a basic agent runtime that executes the core backend pipeline
3. So that a user prompt can be transformed into a structured response based on a defined task schema.

## Acceptance criteria
* The backend provides a basic runtime flow: Prompt -> Intent -> Agent -> LLM -> Response
* The runtime can accept a text prompt as input.
* The intent component may be mocked and maps the input to one predefined example workflow from Schema Foundation #6.
* The selected agent uses the corresponding predefined task schema from #6 .
* The agent sends the prompt to an LLM interface and receives a response.
* The LLM interface may use a mocked response until #18 is completed.
* The runtime returns a structured response object containing at least:
  * selected intent or mocked intent
  * selected agent or workflow
  * LLM response text
  * status or error information
* Basic error handling is implemented for missing schema, LLM failure, or invalid runtime input, especially timeouts
* The implementation is aligned with System Architecture #12.
* The runtime setup and mocked assumptions are documented in the project wiki.

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

shikharthakur2404:
Testing: Added 44 unit tests covering orchestration, intent detection, and type primitives.

Key Changes
   - Core Orchestration: Implemented ConversationOrchestrator to manage the async message pipeline.
   - Intent Detection: Added MockIntentDetector which currently routes to a predefined
     book_service_appointment workflow.
   - Agent Registry: Implemented routing logic to select the correct agent (Sales vs. Technical Support)
     based on the configuration.
   - LLM Integration (Mocked): Integrated a mocked LLM interface with a mandatory 10-second timeout and error
     handling for failures/timeouts.
   - Type Safety & Standards:
       - Structured ConversationResponse with status, error, and intent metadata.
       - Verified full compliance with project standards using ruff (linting/formatting) and mypy
         (type-checking).

Basic foundation
 Prompt -> Intent Detection -> Agent Selection -> LLM -> Response,
 
Ticket/Story/Issue Completed With @djv03 

Thank you

Closes #33


backend/tests/test_orchestration.py
```diff
@@ -50,4 +53,21 @@ async def test_process_message_greets_on_empty_messages() -> None:
...
+
+    # Mock _call_llm to sleep longer than the timeout
+    async def slow_llm(*args: Any, **kwargs: Any) -> str:
+        await asyncio.sleep(2.0)
+        return "too slow"
+
+    with patch.object(ConversationOrchestrator, "_call_llm", side_effect=slow_llm):
+        # Temporarily reduce timeout for the test
+        with patch("asyncio.wait_for", side_effect=asyncio.TimeoutError):
+            response = await orch.process_message(_make_request())
```
K-Vineesh: Heads up: this test triggers a "coroutine was never awaited" RuntimeWarning 
during garbage collection. Reproduce with:

  PYTHONTRACEMALLOC=20 poetry run pytest -W error::RuntimeWarning

Traceback points to this test calling orch.process_message, which constructs a 
coroutine from _call_llm at orchestration/__init__.py:79 that never gets 
awaited. The test still passes because the except TimeoutError branch runs, but 
the timeout pathway isn't actually being exercised end-to-end.

Suggestion: instead of patching asyncio.wait_for (or whatever is short-circuiting 
the await), patch _call_llm itself with an AsyncMock that sleeps past the 
timeout, e.g. AsyncMock(side_effect=lambda *a, **kw: asyncio.sleep(20)). That 
way the real asyncio.wait_for fires the timeout for real.

backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -28,106 +32,136 @@
...
+
+            # 2. Select agent — local import avoids circular dependency with agents/__init__.py
+            from taskorbit.agents import AgentRegistry
+
+            agent = AgentRegistry.get_agent(request.agent_config, self)
+            logger.info(
+                "agent_selected",
+                agent=type(agent).__name__,
+                conversation_id=request.conversation_id,
+            )
```
K-Vineesh, line 77: This is the wiring question from my top-level comment. `agent` is constructed, 
logged, and then dropped; all the downstream methods (_select_active_tool, 
_build_system_prompt, _call_llm) run off `self` and `request.agent_config`. 

Could you either:
(a) confirm in a comment here that the agent object is intentionally a no-op 
    placeholder 
(b) thread the work through the agent (call agent.get_task_definitions() and 
    use that for tool selection)?

Mostly so the next person reading this doesn't think there's a real bug.

backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -28,106 +32,136 @@
...
+
+    async def process_message(self, request: ConversationRequest) -> ConversationResponse:
+        """Main entry point called by the API layer and agent workers."""
+        try:
+            last_user = next(
+                (m for m in reversed(request.messages) if m.role == MessageRole.USER),
+                None,
+            )
+            if not last_user:
+                raise ValueError("No user message found in request.")
```
K-Vineesh: `if not last_user` catches a missing message but not content="". A real LLM 
sent an empty prompt either errors or produces nonsense. One-liner:

    if not last_user or not last_user.content.strip():
        raise ValueError("No user message content found in request.")

backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -28,106 +32,136 @@
...
+                conversation_id=request.conversation_id,
+            )
+
+            # 3. Select active tool
+            active_tool = self._select_active_tool(request.messages, request.agent_config)
+
+            # 4. Build system prompt
+            system_prompt = self._build_system_prompt(request.agent_config, active_tool)
+
+            # 5. Call LLM with a 10-second timeout
```
K-Vineesh: Could we pull this from self._settings if that is possible rather than hardcoding 10.0? Even just 
self._settings.llm_timeout_seconds with a default of 10.0 would make it easy to 
tune when real LLM providers land in #18 (real network calls can have very 
different latency characteristics from the mock's asyncio.sleep(0.5)).

K-Vineesh (commented):
Nice work on this foundation. A few notes below, nothing scary. Main thing is a 
question on the agent-layer wiring; the rest are small.

What works well:
- 44/44 tests pass locally for me, organized cleanly by concern
- Three-tier exception handling (TimeoutError, ValueError, Exception) producing 
  structured ConversationResponse objects is exactly right; the API layer never 
  sees a raw exception
- The module docstring on ConversationOrchestrator states the key invariant 
  clearly ("the LLM only ever sees context for the currently active task"). Good 
  to anchor on for #18
- conversations.py is appropriately thin
- structlog usage is consistent

Main question: the agent layer reads as scaffolding rather than wired-in 
behavior. AgentRegistry.get_agent() runs and the result is logged, but the 
orchestrator then does all the work directly off request.agent_config; the 
SalesAgent and TechnicalSupportAgent objects are never actually invoked. 
get_task_definitions() is defined but never called. The agents' handle_message 
implementations call back into the orchestrator with no guard, which would 
recurse if anything ever did call them.

Is this intentional scope for a follow-up ticket (i.e., the wiring lands when 
real intent routing replaces MockIntentDetector)? If so I think it's fine to 
land this with the TO-DO comments as they are, maybe with a note in the PR 
description or a follow-up issue tracking it. 

Smaller things in inline comments. Happy to approve once those are addressed or 
acknowledged.

djv03:
thanks @K-Vineesh for the keen eyes and inputs

all the issue has been adressed

regarding final the concern you mentioned:

yes — this PR  #33 is scaffolding. The intended future flow is:                                                                      
   
process_message → AgentRegistry → agent.handle_message() → agent owns its task flow                                                                                                   
                                                                                                                                                                                        
But that can't land until MockIntentDetector is replaced by real intent routing #18, because the agent's behavior needs to vary by intent. The TO-DO comments in handle_message() already say this.

qiblatainf:
Getting merge conflicts at the moment

qiblatainf:
Accept incoming changes from main branch

qiblatainf (approved):

backend/tests/test_orchestration.py
shikharthakur2404: >Thanks for catching that! I've updated the test to patch _call_llm
  with a slow side_effect instead of patching asyncio.wait_for. I also
  wired it to use a shorter timeout from Settings during the test, so the
  real wait_for now fires correctly and the coroutine is properly
  awaited/managed. No more RuntimeWarning!

backend/src/taskorbit/orchestration/__init__.py
shikharthakur2404: I've gone with option (a) and added a comment in the orchestrator to
  clarify that the agent object is currently a placeholder. You're right
  that it's not fully wired yet—the plan is to move logic into
  agent.handle_message() once we replace the MockIntentDetector with real
  routing, as the agent's behavior will need to vary based on the
  detected intent. This keeps the foundation clean for now

backend/src/taskorbit/orchestration/__init__.py
shikharthakur2404: Applied! Added the .strip() check as suggested to ensure we don't
  send empty or whitespace-only prompts to the LLM. It now raises a
  ValueError which the orchestrator catches and returns as a structured
  error response

PR #59, opened 2026-05-11
Feature/audio input via livekit

### LINKED ISSUE #14: Audio Input via LiveKit (Speech-to-Text)
## User Story
As an end user  
I want to speak into the microphone and have my speech converted into text  
So that the system can process my spoken input as part of the conversational agent workflow

## Acceptance Criteria
- The frontend requests microphone permission from the browser
- Audio input can be captured through the browser-based frontend
- A simple cutoff mechanism is used (e.g. fixed timeout of ~2 seconds of silence or manual stop)
- Audio is successfully transmitted from the frontend to the backend via LiveKit
- A Speech-to-Text provider is integrated to process incoming audio streams (either ElevenLabs or Deepgram) and evaluated
- The generated transcript is available for further backend processing
- The implementation is aligned with the documented system architecture from System Architecture #12
- The setup is documented for the team in the project wiki, including required .env variables

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

abdulmoeez1225:


backend/src/taskorbit/livekit_agent/llm.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+        if not self._reply_requested:
+            # Preemptive generation call — not triggered by Send. Drop it.
+            return
+        self._reply_requested = False
+
+        messages = _convert_chat_ctx_to_messages(chat_ctx)
+        last_user = next((m for m in reversed(messages) if m.role == MessageRole.USER), None)
+        if last_user:
+            print(f"[STT] {last_user.content}", flush=True)
```
K-Vineesh: BLOCKER. This emits every user transcript to stdout:

    print(f"[STT] {last_user.content}", flush=True)

Looks like a debug artifact. On any hosted environment stdout goes to 
application logs where the content is unbounded and potentially 
sensitive. Replace with the project's structured logger:

    from taskorbit.logging.setup import get_logger
    log = get_logger(__name__)
    ...
    log.debug("stt_transcript_received", content=last_user.content)

If you'd rather not log transcript content at all (probably the right 
call for prod), just log a length:

    log.debug("stt_transcript_received", length=len(last_user.content))

backend/tests/test_livekit_agent_tts.py
K-Vineesh, line 1: BLOCKER. session.py now passes voice_settings=VoiceSettings(...) to the 
ElevenLabs constructor (added in 1d476fb / 13d9e40), but this test still 
asserts the constructor was called only with api_key, voice_id, and 
model. Same issue in the other TTS test in this file and in 
test_livekit_agent.py::test_build_agent_session_uses_deepgram_elevenlabs_silero.

Two ways to fix:

(a) Add voice_settings to the assert_called_once_with(...) so it 
    matches what session.py now passes.

(b) Switch to mock_tts.assert_called_once() and then assert on 
    individual kwargs you care about:

        mock_tts.assert_called_once()
        _, kwargs = mock_tts.call_args
        assert kwargs["api_key"] == "..."
        assert kwargs["voice_id"] == "..."

(b) is more resilient to future constructor additions.

backend/tests/test_worker.py
```diff
@@ -0,0 +1,137 @@
...
+        patch("taskorbit.worker.build_default_agent", return_value=mock_agent),
+    ):
+        await entrypoint(ctx)
+
+    handler = registered["data_received"]
+    packet = _data_packet(json.dumps({"type": "commit_turn"}).encode())
+    handler(packet)
+
+    mock_agent.request_reply.assert_called_once()
+    mock_session.generate_reply.assert_called_once()
```
K-Vineesh: BLOCKER. Race condition:

    handler(packet)                              # schedules a task
    mock_agent.request_reply.assert_called_once()
    mock_session.generate_reply.assert_called_once()  # fails — task hasn't run

The handler calls asyncio.create_task(_commit_and_reply()). That task is 
scheduled but the event loop doesn't run it until control yields. The 
assertion fires before the task ever executes, so generate_reply has 0 
calls.

Fix:

    handler(packet)
    await asyncio.sleep(0)   # yield so the created task runs
    mock_agent.request_reply.assert_called_once()
    mock_session.generate_reply.assert_called_once()

Even better, since _commit_and_reply itself has `await asyncio.sleep(0.3)`, 
you may need `await asyncio.sleep(0.31)` or use freezegun/asyncio time 
mocking. Start with sleep(0) and bump if it's still flaky.

backend/src/taskorbit/worker.py
```diff
@@ -0,0 +1,84 @@
...
+    LIVEKIT_URL, LIVEKIT_API_KEY, LIVEKIT_API_SECRET
+    DEEPGRAM_API_KEY
+    ELEVENLABS_API_KEY
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import logging
```
K-Vineesh: IMPORTANT. Project convention is structlog via 
`from taskorbit.logging.setup import get_logger`. This module bypasses 
that — log lines from worker.py won't carry structured context and will 
look different in any JSON log pipeline.

Replace:
    import logging
    logger = logging.getLogger(__name__)

with:
    from taskorbit.logging.setup import get_logger
    log = get_logger(__name__)

And convert format-string calls to key-value form:
    log.info("worker_generate_reply_triggered")
    log.warning("worker_generate_reply_failed", error=str(exc))

Not a blocker.

K-Vineesh (commented):
Thanks for the depth on this one. 45 files, end-to-end audio pipeline 
wired from browser mic to orchestrator to TTS — that's a lot of moving 
parts working together. I ran the suite locally; verdict below.

What's working well:

- The full chain (Silero VAD → Deepgram STT → orchestrator bridge → 
  ElevenLabs TTS → LiveKit room) is wired end-to-end and the architecture 
  is sound.
- useMicRecorder is well-built: the AudioContext reuse pattern, explicit 
  MediaStreamTrack.stop() to release hardware, RAF-based polling to avoid 
  re-render storms — these are exactly the kind of details that bite in 
  production and they're handled here.
- All three cutoff mechanisms from AC3 are wired: 2s silence detection, 
  manual Pause, manual Send.
- backend/.env.example is thorough — every LiveKit, Deepgram, and 
  ElevenLabs variable is documented with context.

Blocking issues (4 failing tests + 1 debug print). Each has an inline 
comment below:

1. tests/test_livekit_agent_tts.py — 2 tests fail because session.py 
   added voice_settings=VoiceSettings(...) to the ElevenLabs constructor 
   (commits 1d476fb / 13d9e40) but the test assertions weren't updated.
2. tests/test_livekit_agent.py — 1 test fails for the same reason.
3. tests/test_worker.py:98 — race between the synchronous handler() call 
   and the asyncio.create_task() it schedules. Needs an `await 
   asyncio.sleep(0)` to yield to the task before asserting.
4. backend/src/taskorbit/livekit_agent/llm.py:158 — `print(f"[STT] ...")` 
   emits every user utterance to stdout. Almost certainly a debug 
   leftover.

Confirmed locally: `poetry run pytest` reports 4 failed, 49 passed.

Important (not blocking, but worth addressing in this PR):

5. worker.py uses stdlib logging instead of the project's structlog 
   get_logger.
6. The 300ms sleep in _commit_and_reply is a fragile Deepgram-flush 
   heuristic — needs to be tunable at minimum.
7. The data_received handler doesn't filter by participant identity.
8. ConversationalChat.tsx — handleSegment depends on the whole `call` 
   object, which changes identity on every render. This re-registers 
   the LiveKit text stream handler repeatedly and can drop in-flight 
   transcripts.

A couple of open questions rather than findings:

9. STT provider rationale — the acceptance criterion says "evaluated." 
   Could you add a sentence to the PR body or a commit message explaining 
   why Deepgram over ElevenLabs STT (or other alternatives)? Even a 
   one-liner ("lower latency in our test", "better German support", 
   etc.) satisfies the AC.
10. Does livekit-agents' AgentSession handle teardown automatically on 
    room disconnect for the SDK version you're using? If not, we'll 
    want a try/finally with session.aclose() around session.start(). 
    Happy to take this as a follow-up if the SDK handles it.

Smaller polish items in inline nits.

K-Vineesh (approved):

PR #64, opened 2026-05-13
feat: User Conversation History persist and retrieve conversations

1Endri:
## What was done
- Added POST /v1/conversations - create a new conversation automatically
- Added GET /v1/conversations - get all conversations
- Added GET /v1/conversations/{conversation_id}/messages - get all messages for a conversation
- Modified POST /v1/conversations/process - now auto-saves user and assistant messages to DB on every turn
- Conversations are uniquely identified by conversation_id
- Added 3 new tests for the new endpoints (5 total in conversations route tests)
- All 111 tests passing

## Acceptance Criteria Coverage
- Conversation is automatically created when first message is sent via /process
- All user and system messages are stored in the backend
- Conversations are uniquely identified by conversation_id
- Reloading restores previous conversations - backend endpoints ready, frontend integration pending (in progress)

## Notes
- Structure is multi-user ready (user_id as nullable FK on conversation_messages)
- Frontend integration (restore on reload) requires coordination with frontend dev
- Wiki documentation will be updated after review

## How to test
docker compose up -d postgres
cd backend
poetry run alembic upgrade head
poetry run pytest tests/test_conversations_route.py -v


backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -21,12 +35,110 @@ def get_orchestrator(
...
     try:
-        # process the message through the orchestration engine and return the response
-        return await orchestrator.process_message(request)
+        response = await orchestrator.process_message(request)
+
+        # Save user message
+        last_user = next(
+            (m for m in reversed(request.messages) if m.role == MessageRole.USER),
+            None,
+        )
```
abdulmoeez1225: The /process endpoint finds the last user message by scanning backward through request.messages, which is the entire conversation history. On turn 2, request.messages = [user_1, assistant_1, user_2]. The code
  finds user_2 — correct. But if the client ever resends history without the latest message appended last, it will re-save an already-persisted message. The intent should be: save only request.messages[-1] if it's
   a user role, not search backward.

  # Current (fragile)
  last_user = next(
      (m for m in reversed(request.messages) if m.role == MessageRole.USER),
      None,
  )

  # Simpler and correct
  last_msg = request.messages[-1] if request.messages else None
  last_user = last_msg if last_msg and last_msg.role == MessageRole.USER else None

abdulmoeez1225 (commented):
## Code Review

### Overview

This PR adds conversation persistence: three new REST endpoints (`POST/GET /v1/conversations`, `GET /v1/conversations/{id}/messages`), auto-saving messages on `/process`, a full async migration of the CRUD layer, and a frontend card showing prior conversation count on page load.

---

### 🔴 Critical Issues

**1. Message double-saving risk (`conversations.py:94–103`)**

The `/process` endpoint finds the last user message by scanning backward through `request.messages`, which is the entire conversation history. On turn 2, `request.messages = [user_1, assistant_1, user_2]` works fine — but if the client ever resends history without the latest message appended last, it will re-save an already-persisted message. The intent should be: save only the last element if it is a user role.

```python
# Current (fragile)
last_user = next(
    (m for m in reversed(request.messages) if m.role == MessageRole.USER), None,
)

# Simpler and correct
last_msg = request.messages[-1] if request.messages else None
last_user = last_msg if last_msg and last_msg.role == MessageRole.USER else None
```

---

**2. Silent failure on message persist (`conversations.py:98–113`)**

`create_conversation_message` returns `None` on SQLAlchemy error and swallows the exception. The route receives `None` and continues, returning HTTP 200 even though the message was never saved. Either raise on failure or at minimum log a warning when the return value is `None`.

---

**3. Orphaned messages risk (`conversations.py:98–113`)**

`/process` saves messages with `request.conversation_id` but never checks that a `Conversation` row with that ID exists first. With PostgreSQL FK enforcement this raises an IntegrityError (caught and swallowed by `crud.py` → returns `None`). Messages can silently be saved under nonexistent conversation IDs in production.

---

### 🟡 Significant Issues

**4. Type mismatch between frontend and backend (`conversationApi.ts:87–96`)**

The `ConversationsResponse` type declares `agent_id` on each conversation item and a top-level `total` field — neither of which the backend actually returns. The GET route only returns `id`, `agent_name`, `started_at`, `ended_at`.

---

**5. Raw DB access in route handlers (`conversations.py:128–143`, `152–167`)**

`create_conversation` and `get_conversations` execute SQL directly in the route handlers, bypassing the CRUD layer. Add `create_conversation` and `get_all_conversations` to `crud.py` for consistency with the rest of the codebase.

---

**6. Loose return types (`conversations.py:127`, `149`, `174`)**

All three new endpoints return `-> dict`. This suppresses OpenAPI schema generation. Use a Pydantic `BaseModel` response class.

---

**7. `console.log` in production frontend (`ConversationalChat.tsx:930–933`)**

```ts
console.log("Loaded conversations from DB:", data);
console.log(`Found ${data.conversations.length} previous conversations`);
```

Remove before merge.

---

**8. Loose state type (`ConversationalChat.tsx:919`)**

```ts
// Current
useState<Record<string, unknown>[]>([])

// Should be
useState<ConversationsResponse['conversations']>([])
```

`ConversationsResponse` is already defined in `conversationApi.ts`.

---

### 🔵 Minor Issues

**9.** No pagination on `GET /v1/conversations` — add a default limit matching the pattern in `get_messages_by_conversation`.

**10.** `agent_id="default"` is hardcoded (`conversations.py:131`) — meaningless data in the DB; should come from the request body.

**11.** `GET /v1/conversations` returns all conversations across all users (acknowledged in PR). The nullable `user_id` FK means there is no way to filter by caller identity. Needs a plan before any real deployment.

**12.** `_normalize_text` is in `livekit_agent/llm.py` (LiveKit path) but the `UnicodeEncodeError` catch is in `orchestration/__init__.py` (REST path). If normalization already runs before orchestration, the catch is dead code. Clarify whether both paths need normalization.

---

### 🧪 Test Coverage Gaps

- `test_create_conversation_returns_201`: `_mock_db()` sets up `db.execute` but the create path uses `db.add/commit/refresh`. The test passes only because `AsyncMock` auto-creates all attribute access — it does not verify a conversation was actually created.
- No tests for error paths (500 responses from the three new endpoints).
- No test verifying `/process` calls `create_conversation_message` with the correct role and content.
- No test for empty `request.messages` or messages with no user turn.

---

### ✅ What Works Well

- The async CRUD migration is clean and consistent — every function correctly awaits, uses `select()` with `AsyncSession`, and the `aiosqlite` test fixture is the right approach.
- `_normalize_text` with a clear explanatory comment is well-done.
- `_mock_db()` is a good reusable test helper.
- The `UnicodeEncodeError` handler in orchestration is a reasonable defensive addition.

---

### Summary

The core persistence logic is directionally correct, but the message-saving path in `/process` has real data integrity bugs (silent failures + FK orphaning), and the frontend fetch is wired up but does nothing with the data beyond showing a count. Please fix the critical issues before merging, and track the auth gap as a required follow-up.

abdulmoeez1225 (approved):

PR #79, opened 2026-05-20
fix: auto-create conversation row in /process to prevent FK violation

### LINKED ISSUE #45: User Conversation History

1Endri:
Fixes #45 regression reported during Sprint 5 RC smoke test.

## Problem
The frontend generates a conversation_id client-side and posts directly 
to /v1/conversations/process without first calling POST /v1/conversations 
to create the parent row. This caused a ForeignKeyViolation because 
conversation_messages.conversation_id FK had no parent row to reference.
Messages were silently dropped and the history tab stayed empty.

## Fix
Auto-create the Conversation row inside /process if it does not exist 
before saving messages. Uses the agent_id and agent_name from the 
request payload instead of hardcoded defaults.

## Testing
- All 111 tests passing
- Linting clean
- Manually verified: messages now persist correctly after fix


backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -65,7 +77,10 @@ async def process_conversation(
...
             )
             if saved is None:
-                logger.error("failed_to_save_user_message", conversation_id=request.conversation_id)
+                logger.error(
+                    "failed_to_save_user_message",
+                    conversation_id=request.conversation_id,
+                )
 
         # Save assistant reply
         if response.reply:
```
Asadraza17: Edge case: the assistant-message save at line 86 is outside the `if last_user:` block. If a request arrives with no user message (system-only messages, or an empty messages array with a generated reply), the auto-create on lines 54-71 is skipped — but this block still tries to INSERT an assistant message with that conversation_id. That'll hit the same FK violation this PR is supposed to fix.

Move the `if response.reply:` block inside the `if last_user:` guard, OR move the auto-create out of the user-guard so it runs unconditionally when persistence is attempted.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -51,12 +51,24 @@ async def process_conversation(
...
             if not conversation:
-                logger.warning("conversation_not_found", conversation_id=request.conversation_id)
+                conversation = Conversation(
+                    id=request.conversation_id,
+                    agent_id=request.agent_config.id,
+                    agent_name=request.agent_config.name,
+                    started_at=datetime.now(UTC),
+                )
+                db.add(conversation)
+                await db.commit()
```
Asadraza17: This commits the conversation row immediately. If the subsequent message save fails (LLM timeout mid-handler, DB hiccup), you end up with an orphan conversation row with no messages — or a partial conversation with only the user side, if the assistant save specifically fails.

Strict reading of AC "all user and system messages are stored in the backend": under partial failure, "all" becomes "some."

Suggestion: wrap the conversation auto-create + user message + assistant message into a single transaction (add to session, commit once at the end). The crud functions currently each commit internally — you'd want a non-committing variant or one batched helper.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -51,12 +51,24 @@ async def process_conversation(
         last_user = last_msg if last_msg and last_msg.role == MessageRole.USER else None
```
Asadraza17: DoD #4: "Unit tests for new features have been written" — but the diff adds zero test files.

The 174 existing tests still pass, but they cover what main already had. Nothing tests the new behavior:
- auto-create branch when conversation doesn't exist
- idempotent reuse when it already exists
- agent_id/agent_name copied correctly from request
- no-user-message edge case (related to comment above)

Would suggest 3-4 tests in backend/tests/test_conversations_route.py

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -51,12 +51,24 @@ async def process_conversation(
...
 
         if last_user:
```
Asadraza17: DoD #3: "Documentation is updated (System Architecture, SBOM)."

This PR doesn't touch any .md file. The new conversation_auto_created event, the History endpoints flow, and the FE wiring should be reflected in the wiki — either the System-Architecture page or a new entry on the LLM-Integration or Database-Setup pages.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -51,12 +51,24 @@ async def process_conversation(
...
+            # Auto-create conversation if it doesn't exist (FK safety)
```
Asadraza17: Branch name says fk-auto-create, but the PR also bundles the History page FE rewiring (commit fdd5ba8) which is the Sprint 6 #45 carryover work — separate from the Sprint 5 hotfix. These could/should have been two PRs.

Not blocking the merge, but flagging for future scope discipline.

frontend/src/pages/HistoryPage.tsx
```diff
@@ -1,92 +1,69 @@
...
-  const m = Math.floor(seconds / 60);
-  const s = seconds % 60;
-  return m > 0 ? `${m}m ${s}s` : `${s}s`;
-}
+type Conversation = {
+  id: string;
+  agent_name: string;
+  started_at: string;
+  ended_at: string | null;
+};
```
Asadraza17, line 17: This `Message` type duplicates the shape inline-typed inside getConversationMessages' return in conversationApi.ts. Worth exporting a `ConversationMessage` type from conversationApi.ts and importing here, so the source of truth lives in the API client.

frontend/src/pages/HistoryPage.tsx
```diff
@@ -1,92 +1,69 @@
-import { Fragment, useEffect, useRef, useState } from "react";
-import { DatabaseZap, MessageSquareDashed } from "lucide-react";
+import { useEffect, useRef, useState } from "react";
```
Asadraza17: The new HistoryPage drops several features from main:
- "Today, 14:30" / "Yesterday, 14:30" relative date formatting
- Duration pill ("3m 42s")
- Language badge
- Extracted-data sidebar
- TranscriptBubble + TranscriptToolMarker components
- Tool firings rendering

Some of these can't be supported until the backend exposes the underlying data (duration, language, tool firings) — but the PR doesn't acknowledge that the History page is now functionally less featured than before.

Could you:
Keep the rendering components and just feed them null/defaults from the backend response?

Asadraza17 (commented):
Did a deep validation pass. Text-chat path through POST /v1/conversations/process works as advertised — auto-create fires, both messages persist, idempotency holds, History endpoints return real data. Verified via curl + DBeaver.

But this PR doesn't fix the bug for the actual user-facing demo path: the voice chat surface.

Reproducer (just ran this on the branch):
1. docker compose up
2. Open localhost:5173
3. Start a voice session, talk to John Doe, end the call
4. Open /history → empty list
5. DBeaver → conversations table → zero new rows

The voice path goes through the LiveKit worker (backend/src/taskorbit/livekit_agent/llm.py), which calls ConversationOrchestrator.process_message() directly (line 203), NOT the FastAPI route where the auto-create lives. Worker logs confirm: conversation_id=livekit-session, hardcoded default from line 141.

Two underlying issues that make voice persistence broken end-to-end:

(a) The auto-create logic is only in the FastAPI route. The orchestrator method has zero persistence code (grep -E "db\.add|db\.commit" backend/src/taskorbit/orchestration/__init__.py returns empty). The worker calls process_message() directly, bypassing the route entirely.

(b) The FE never pipes its generated conversationId to the worker. useVoiceCall.ts:72 generates conv-<timestamp>-<random>, but buildLiveKitWorkerMetadata in livekitAgentMetadata.ts doesn't include conversation_id, so the worker falls back to the hardcoded "livekit-session" — meaning even if persistence were wired into the worker, every voice call would collide on the same ID, violating AC "Conversations are uniquely id'd."

Requesting changes — this needs to be fixed before merging because the demo path for the mid-project review is voice, not text.

Possible fixes (your call which way):
- Move persistence into ConversationOrchestrator.process_message() so both routes and the worker benefit (DRY, fixes both at once)
- OR keep persistence in the route AND have the worker call the same route via HTTP (more network hops, less ideal)
- AND in either case: pipe FE conversationId through the LiveKit JWT metadata into the worker's OrchestratorAgent constructor

Happy to re-validate once the voice path persists.

backend/tests/test_conversations_route.py
```diff
@@ -110,3 +111,103 @@ def test_create_conversation_returns_201() -> None:
         response = client.post("/v1/conversations")
     assert response.status_code == 201
     app.dependency_overrides = {}
+
+
+# ============ MOCK-BASED TESTS FOR AUTO-CREATE CONVERSATION BEHAVIOR ============
+
+
+def test_auto_create_conversation_logic_when_not_exists():
```
Asadraza17: DoD #4 (unit tests for new features) — the 4 new tests here don't exercise the new code path. They construct a Conversation(...) object inside the test body and assert its own attributes — no TestClient call, no process_conversation call, no DB session, no mock of create_conversation_message. Coverage of the auto-create branch you added is 0%.

What's needed (use the existing [test_process_conversation_returns_200_with_mock](vscode-webview://08nbo5qi4m38th6hrhkoq83n0cr6qej3gjcb3rd2udj784unhjn5/backend/tests/test_conversations_route.py#L41) pattern — app.dependency_overrides + patch("...create_conversation_message") is already set up correctly):

Auto-create fires when conversation is missing — POST /v1/conversations/process with a fresh conversation_id, mock db.execute to return None for the lookup, assert db.add was called once with a Conversation whose agent_id/agent_name match the request.
Idempotent when conversation exists — mock the lookup to return an existing Conversation, POST, assert db.add was NOT called.
Agent fields come from the request, not hardcoded — POST with a custom agent_config.id / name, assert the auto-created row has those exact values (not "default" / "TaskOrbit" from the create_conversation route's hardcoded defaults).
No-user-message edge case — POST with messages array that has no role == "user" entry, assert no auto-create + no create_conversation_message call.
The current 4 tests assert that Python attribute assignment works — that's not the same as testing the route.

backend/src/taskorbit/api/routes/conversations.py
Asadraza17: You added process_message(request, db=None) with _persist_messages in the orchestrator (good design), but this route at line 47 still calls await orchestrator.process_message(request) without passing db=db. So _persist_messages is dead code on the REST path — all the persistence here at lines 53-100 duplicates what the orchestrator method would do.

Pick one path:

(preferred) Pass db=db to process_message and delete lines 53-100. Orchestrator handles persistence for both REST + voice via one wrapped transaction. Single source of truth.
Or delete _persist_messages from the orchestrator and keep the route as-is.
Two paths drift apart over time — already true here: orchestrator wraps in async with db.begin():, route does not.

Asadraza17: DoD #3 (documentation updated — System Architecture / SBOM) - diff doesn't touch any .md or wiki page. Doesn't need to be long, but needs:

Short wiki entry under System Architecture: the auto-create-on-first-message contract + the two new History endpoints (GET /v1/conversations, GET /v1/conversations/{id}/messages) + how FE wires to them.
SBOM: no new deps in this PR (verified pyproject.toml / package.json untouched), so a one-line "no SBOM impact" note is fine.

Asadraza17 (commented):
Round-2 pass - re-graded everything strictly against issue #45 AC + DoD. Three blockers left, one soft note, all inside scope. Detailed below.

backend/src/taskorbit/livekit_agent/__init__.py
```diff
@@ -8,11 +8,95 @@
...
+    )
+
+
+def build_default_agent(settings: Any = None) -> OrchestratorAgent:
+    """Build a default agent for testing without a database session."""
+    orchestrator = _get_orchestrator()
+    return OrchestratorAgent(
+        orchestrator=orchestrator,
+        conversation_id=str(uuid.uuid4()),
+        db=None,
```
Asadraza17, line 73: Voice persistence is still broken end-to-end (Comment 7 from round 2).

The orchestrator now persists when given a db session, but here it's hardcoded db=None. worker.py calls build_default_agent(settings=cfg) directly, so every voice call goes through process_message with db=None, which short-circuits the new _persist_messages (if db is not None: skips it). Net result: voice path has zero persistence. AC "All user and system messages are stored in the backend" still violated on the demo path for the mid-project review.

Also: conversation_id=str(uuid.uuid4()) here means the FE's conv-<uuid> from useVoiceCall.ts:72 is still discarded. And buildLiveKitWorkerMetadata in frontend/src/lib/livekitAgentMetadata.ts still doesn't include conversation_id in the JWT metadata — I checked.

Two pieces to land this fully:

In the worker entrypoint, open an AsyncSession per call (or per turn — async_session_maker() from database.session) and pass it here: OrchestratorAgent(db=session).
Pipe conversation_id through the LiveKit JWT metadata → participant.metadata → OrchestratorAgent(conversation_id=...) so voice calls don't all collide on the same UUID and persist under the same id the FE expects to load in the History page.

backend/src/taskorbit/api/routes/conversations.py
Asadraza17, line 1: Documentation still missing (DoD #3 — Comment 11 from round 2).

Diff doesn't touch any .md or wiki page. The DoD says "Documentation is updated (system architecture, SBOM)."

Doesn't need to be long, just needs:

Short wiki entry under System Architecture: the auto-create-on-first-message contract + the two new History endpoints (GET /v1/conversations, GET /v1/conversations/{id}/messages) + how FE wires to them.
SBOM: no new deps in this PR (verified pyproject.toml / package.json untouched), so all good there.

Asadraza17 (commented):
Round-3 check done. Two blockers from round 2 are still open, though. Would flip to Approve once voice persistence actually fires end-to-end on the demo path + the wiki entry lands.

Asadraza17 (approved):
Round-5 verified end-to-end. Both user + assistant messages persist now (DB check: symmetric counts across 3 test conversations), voice path also persisting correctly. Approving, good to merge.

Signed-off-by: Asad Raza <asadraza1702@gmail.com>

PR #92, opened 2026-05-25
Feature/70 slot extraction

djv03:


backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -94,6 +109,17 @@ async def process_message(self, request: ConversationRequest) -> ConversationRes
             _llm_elapsed = time.perf_counter() - _llm_start
             get_metrics().pipeline_latency_seconds.labels(stage="llm_call").observe(_llm_elapsed)
 
+            # 5b. Execute DataExtractionTool when all required slots are filled
+            if slot_result.is_complete and intent.required_inputs:
+                from taskorbit.tools.data_extraction import DataExtractionTool
+
+                await DataExtractionTool().execute(slot_result.to_dict())
```
K-Vineesh, line 116: The return value of DataExtractionTool.execute() is not captured here. 
If execution fails, there is no log, metric, or signal to the caller. 
At minimum:

result = await DataExtractionTool().execute(slot_result.to_dict())
if not result.success:
    logger.warning("data_extraction_tool_failed", error=result.error)

Not a blocker for this PR but worth tracking as a follow-up.

K-Vineesh (approved):
AC1 through AC7 are met for this sprint scope. Slot extraction is wired 
end-to-end, missing slots are detected and surfaced in the prompt, and 
multiple data types are covered.

One inline comment on line 116 — DataExtractionTool result is silently 
dropped on failure. Not blocking merge but should be a follow-up ticket.

PR #127, opened 2026-06-05
Feature/tools end call

### LINKED ISSUE #116: [Tools] End Call
## User story
1. As a user
2. I want the agent to end call if I ask him to
3. So that I can end the conversation through my voice input

## Acceptance criteria
* A new tool definition is created for the new agent functionality.
* The new tool can be found in the drop down menu in the Agent Config page.
* A user can select the tool for any given agent definition.
* An active agent that includes this tool in It's config is capable of ending the call if asked by the user.

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

qiblatainf:


Asadraza17 (approved):
Approving! 👍

### Verified

- AC1-4 all satisfied. The orchestrator early-exit + worker data-channel publish + FE `useSessionEnded` hook form a clean, symmetric pattern across text and voice paths.
- 77 backend tests pass on local checkout, ruff check + format clean, FE typecheck + prettier clean on all 7 changed files.
- The missing-description bug fix is a great catch, sticking `description: str = ""` plus the targeted regression test in `test_types.py` documents the root cause clearly.
- Keyword list is comprehensive (40+ phrases across 5 semantic categories) with both positive and negative parametrized tests.

### Non-blocking observations

1. **LLM farewell has no fallback.** If `_call_llm` times out or errors during farewell generation in the early-exit path, the user gets an error reply instead of a hangup. Worth wrapping in a try/except with a hardcoded `"Goodbye!"` fallback so user-initiated end-call always succeeds.
2. **`end_call_check` debug log dumps a 60-char message snippet** (`orchestration/__init__.py:73-81`). Low risk because DEBUG is off in prod, but worth dropping the `message_snippet` field if DEBUG ever gets enabled aggregation-side.
3. **No docs update.** The DoD lists "Documentation is updated" - a short `Documentation/end-call-tool.md` or wiki page listing which phrases trigger end-call would help admins. Happy to land as a follow-up.
4. Substring matching on signals like `"end the call"` will also match `"please don't end the call"` - again, low risk given the chosen phrases, just flagging.

PR #131, opened 2026-06-09
Feature/49 critical action confirmation

### LINKED ISSUE #49: Critical Action Confirmation
## User story
1. As a user
2. I want to confirm or reject critical actions before they are executed  
3. So that I stay in control and avoid unintended operations.

## Acceptance criteria
* System can return a "confirmation required" state for specific actions  
* The confirmation request is displayed in the UI with clear context (what will happen)  
* User can explicitly confirm or reject the action  
* User decision is sent back to the backend  
* On confirmation, the system continues execution  
* On rejection, the system aborts the action  
* The result (executed or aborted) is visible in the chat UI  
* End-to-end flow works: Backend → FE → User → Backend → Result

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

djv03:
## Summary
   - Backend: pause orchestration for critical actions and return confirmation payload
   - Frontend: wire confirmation_id/decision flow across types, API, ConfirmationPrompt, useVoiceCall, and ConversationalChat
   - Both chat and LiveKit voice pipelines now respect agent ConfirmationsConfig instead of hardcoding confirmation.required

   ## Test plan
   - [ ] Trigger complaint intent in chat, fill all slots, confirm prompt appears
   - [ ] Click Approve — agent proceeds
   - [ ] Click Deny — agent cancels with message
   - [ ] Verify same flow works in voice pipeline

   Closes #49


1Endri:
Reviewed locally all automated checks pass on my end:
- Backend: 355 passed, 0 failed (3 pre-existing async mock warnings, unrelated to this PR)
- Backend lint (ruff): clean
- Frontend lint: 0 errors (1 pre-existing react-refresh warning in active-agent-provider.tsx, not introduced here)
- Frontend type-check: clean

The implementation is solid and the 3-step flow test (pending → approved → rejected) in test_orchestration.py is well structured. Nice work.

A few things worth confirming / noting before or shortly after merge:

**1. Default confirmation gate (please confirm before merge)**
toolNeedsConfirmation() defaults to required=true when an agent has no confirmations config. This is a safe default, but it means any existing agent config without an explicit confirmations field will now prompt before every tool execution. Can we confirm the shipped/deployed agent templates include an explicit confirmations config so this doesn't unintentionally gate every tool in production?

**2. Voice path has no approve/deny UI (suggest follow-up issue)**
livekitAgentMetadata.ts now correctly sends confirmation.required=true for the voice path (previously hardcoded false), but there's no voice-side UI for the user to approve/deny mid-call — the confirmation flow currently only works through the text chat path. Suggest tracking this as a follow-up issue.

**3. REJECTED status not explicitly handled on frontend (minor)**
The ConversationResponse type now includes "rejected", but ConversationalChat doesn't branch on it explicitly. It works in practice because the backend sends a cancellation reply that gets appended as a normal assistant turn, so the status itself is silently ignored. Not a blocker, just noting it.

**4. No confirmation dedup guard (minor, unlikely)**
If the backend ever returns confirmation_required twice for the same confirmation_id, the UI would call triggerConfirmation again and overwrite pendingConfirmationIdRef. Unlikely in practice but there's no dedup check.

Overall this looks good to me. My only real ask before merge is confirming point 1; the rest can be follow-ups.

shikharthakur2404:
I have updated the implementation to ensure that existing agent configurations are not impacted by this
  change:
  > 
  > 1.  Fallback Logic: Changed toolNeedsConfirmation (in both text and voice paths) to return false if the
  confirmations object is missing. This ensures backward compatibility for all existing agents.
  > 2.  Updated Seed Data: Added explicit confirmations settings to the default agent templates in
  seed_data.py, initialized to required: false.
  > 3.  UI Toggle Fix: This adjustment also resolves a mapping issue where the 'Require confirmations' toggle
  in the editor was non-functional because both the disabled state (undefined) and the previous default mapped
  to true.
  
  @1Endri

1Endri (approved):
Re-tested after c6c7f5c all checks pass (355 backend tests, ruff, lint, type-check all green). 
Confirmed toolNeedsConfirmation now defaults to false when confirmations is undefined in both 
conversationApi.ts and livekitAgentMetadata.ts, so legacy agents are unaffected. Thanks for 
addressing the default gate concern. Approving 👍

Issue #66, opened 2026-05-15
Generic Adapter for external Tools


Caerum5:
## User story
1. As a system admin
2. I want a generic API adapter and configuration interface
3. So that I can easily expand the agent's capabilities with new tools without having to write custom code

## Acceptance criteria
* A generic adapter interface exists that allows defining external tools via configuration (e.g. URL, method, headers, parameters) and assigning them to specific agents, both without requiring custom coding
* The adapter exposes a standardized request and response schema independent of the underlying tool configuration that can be used by the agent runtime to register the assigned tools with the LLM
* The adapter defines and outputs a standardized error response format prepared for future tool call failures (e.g., timeouts)
* The adapter design allows adding new tools without modifying the core adapter implementation
* The adapter's functionality is successfully tested using at least one external dummy tool

## Definition of done (DoD)
* General coding standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

Issue #55, opened 2026-05-08
Voice Interruption for Audio Chat


Caerum5:
## User story
1. As an user
2. I want the voice agent to stop speaking immediately when I start talking
3. So that I can provide new information or corrections without having to wait for the agent to finish the response.

## Acceptance criteria
* The system must detect user voice activity while the agent is currently playing audio back to the user
* After detecting user speech, the agent's current audio playback must be interrupted with a minimal delay
* Any ongoing LLM generation or TTS tasks for the interrupted message must be cancelled to save resources and reduce latency for the new request
* Little background noises should not trigger an such a interruption (e.g. through thresholds)

## Definition of done (DoD)
* General coding standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

PR #132, opened 2026-06-09
Feature/57 stream llm responses

### LINKED ISSUE #57: Stream LLM Responses
## User story
1. As an user
2. I want the agent to start speaking almost immediately after I finish asking a question
3. So that the conversation feels completely natural without pauses and reduce latency

## Acceptance criteria
* The backend must stream the LLM text response token by token (or in small chunks from LLM) instead of waiting for the full generation to complete 
* The system must feed this live incoming text stream to TTS engine directly
* The TTS engine must begin processing and playing audio as soon as the first parts (the first phrases or sentences etc.) of the LLM response are ready
* The resulting voice output should play continously to the user while the LLM is still generating the rest of the text in the background
* The LLM Stream can be aborted midway if an interruption signal is received (e.g., triggered by the Voice Interruption #55)

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

djv03:
## Summary
   - Added `generate_stream()` to LLM client protocol for both OpenAI and Gemini providers
   - Extracted `_prepare_reply` and added `process_message_stream` to orchestration for SSE-based streaming
   - Added `POST /v1/conversations/stream` SSE endpoint streaming tokens as `text/event-stream`
   - Wired LLM chunk streaming through `llm_node` to LiveKit TTS for voice pipeline
   - Frontend: added `sendMessageStream` SSE helper and wired it into `ConversationalChat.handleSendText`

   ## Test plan
   - [ ] Send a message in chat — response should stream in token by token instead of appearing all at once
   - [ ] Verify voice pipeline streams LLM chunks through to TTS without buffering the full response
   - [ ] Check SSE endpoint directly via `POST /v1/conversations/stream` returns `text/event-stream`
   - [ ] Confirm existing non-streaming `/process` endpoint still works unchanged

   Closes #57


backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -424,6 +460,171 @@ async def process_message(self, request: ConversationRequest) -> ConversationRes
...
+            if not llm_text:
+                raise LLMConfigError("LLM returned an empty streaming response")
+
+            # Step 5b: Dispatch active tool when slots are complete
+            from taskorbit.types import ConversationStatus, ToolType
+
+            tool_data: dict[str, Any] = {}
+            response_status = ConversationStatus.SUCCESS
+
+            if prep.active_tool and prep.slot_result.is_complete and prep.intent.required_inputs:
```
Asadraza17: This stream dispatch never replicates the #49 confirmation gate that process_message has: it doesn't check prep.active_tool.confirmation.required, never reads request.confirmation_id/decision (both exist on ConversationRequest), and dispatches unconditionally once slots are complete. A confirmation-required tool's execute() fires with no user consent on the first streamed turn, and CONFIRMATION_REQUIRED is never emitted so the FE can't prompt. Please port the confirmation short-circuit here, or route confirmation-gated turns through POST /process until streaming-confirmation is designed.

backend/src/taskorbit/orchestration/__init__.py
```diff
@@ -424,6 +460,171 @@ async def process_message(self, request: ConversationRequest) -> ConversationRes
...
+                dispatch_context: dict[str, Any] = dict(prep.slot_result.to_dict())
+                if prep.active_tool.type == ToolType.AGENT_TRANSFER:
```
Asadraza17: Three divergences from process_message's dispatch (269-345): (1) no EXTERNAL_API branch, so a #66 GenericApiTool gets dict(slot_result) instead of {**tool.parameters, 'args': slots} and will misbehave; (2) the guard requires prep.intent.required_inputs truthy, so an END_CALL/AGENT_TRANSFER tool with no required slots never fires; (3) no reject-decision path. Recommend factoring the shared dispatch into one helper called by both paths so they can't drift again.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -135,6 +137,95 @@ async def process_conversation(
...
+
+    if meta.status == "error":
+        yield f"data: {json.dumps({'type': 'error', 'message': meta.error or 'Unknown error'})}\n\n"
+        return
+
+    # Persist messages only after a successful full stream
+    last_msg = request.messages[-1] if request.messages else None
+    last_user = last_msg if last_msg and last_msg.role == MessageRole.USER else None
+    if last_user:
+        saved = await create_conversation_message(
```
Asadraza17: Unlike /process (52-63), the stream path never auto-creates the conversation, so this create_conversation_message runs against a conversation_id that isn't in the DB yet. conversation_messages.conversation_id is a NOT NULL FK, the INSERT violates it, crud swallows the SQLAlchemyError and returns None, and the turn is silently lost. The FE generates the id client-side via call.start(), so this is the normal first-turn case. Mirror /process: get_conversation, create if absent, and set request.conversation_id before persisting.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -135,6 +137,95 @@ async def process_conversation(
...
+    Event schema (agreed with frontend / Dev 2):
+      data: {"type": "chunk", "text": "<token>"}
+      data: {"type": "done", "intent": "...", "status": "...", "selected_agent": "...",
+              "slots": {...}, "missing_slots": [...], "conversation_id": "..."}
+      data: {"type": "error", "message": "..."}
+    """
+    meta: ConversationResponse | None = None
+
+    async for event in orchestrator.process_message_stream(request):
+        if await http_request.is_disconnected():
```
Asadraza17, line 178: On disconnect the generator returns before any persistence, so neither the user utterance nor partial assistant text is saved. /process saves the user message up front so a turn survives a failed reply. For the barge-in flow this PR targets (#57, #55), erasing the user's words on every interruption may be wrong. Intended? If user turns should survive interruption, persist the user message up front like /process; if intended, a one-line comment would stop this reading as a bug.

backend/src/taskorbit/livekit_agent/llm.py
```diff
@@ -239,7 +239,20 @@ async def llm_node(  # type: ignore[override]
...
             current_intent_name=self._locked_intent_name,
         )
-        response = await self._orchestrator.process_message(request)
+        from taskorbit.types import ConversationResponse as _ConversationResponse
+        from taskorbit.types import ToolType as _ToolType
+
+        response: _ConversationResponse | None = None
+
+        async for event in self._orchestrator.process_message_stream(request):
+            if isinstance(event, str):
```
Asadraza17: Only str events reach TTS here, so an error-only ConversationResponse (the configured/runtime error branches in process_message_stream) is captured into response but never spoken: the user hears silence. On main the error reply.content was the single yielded chunk, so this is a regression. After the loop, if no chunk was yielded and response.reply.content is set, fall back to yield response.reply.content before persistence.

backend/src/taskorbit/integrations/llm/openai_client.py
```diff
@@ -146,3 +147,106 @@ async def generate(
...
+                provider="openai",
+                model=llm_config.model,
+                chars=char_count,
+                prompt_tokens=prompt_tokens,
+                completion_tokens=completion_tokens,
+                llm_api_latency_ms=round(_api_elapsed * 1000, 1),
+            )
+            _m = get_metrics()
+            _m.llm_requests_total.labels(
+                provider="openai", model=llm_config.model, status="success"
```
Asadraza17: This status="success" is in the finally, so it fires even when the in-loop except openai.APIError re-raised or the generator was closed early by a barge-in: a failed/interrupted stream gets counted as both failure and success. Gate it on a completed = True flag set after the async for, or move the success counter onto the normal-completion path. Same fix in gemini_client.py.

backend/tests/test_orchestration.py
```diff
@@ -609,6 +609,111 @@ async def test_process_message_confirmation_flow(mock_good_intent: Any) -> None:
...
+# ---------------------------------------------------------------------------
+
+
+async def _fake_llm_chunks(*tokens: str):
+    for token in tokens:
+        yield token
+
+
+@pytest.mark.asyncio
+async def test_process_message_stream_yields_chunks_then_response(
```
Asadraza17, line 634: The stream tests cover only chunks-then-response, metadata, empty-messages, and clarification. There's no streaming test for tool dispatch, confirmation_required in the stream path (which would have caught the missing gate above), or a provider that yields a few chunks then raises. Since the two paths have already diverged on exactly these branches, please add at least a stream confirmation-required test asserting _dispatch_tool is NOT called, and a mid-stream-error test.

backend/tests/test_stream_route.py
```diff
@@ -0,0 +1,194 @@
...
+        with TestClient(app) as client:
+            response = client.post("/v1/conversations/stream", json=_VALID_PAYLOAD)
+
+    events = _parse_sse_events(response.text)
+    done_events = [e for e in events if e["type"] == "done"]
+
+    assert len(done_events) == 1
+    done = done_events[0]
+    assert done["status"] == "success"
+    assert done["conversation_id"] == "conv-1"
```
Asadraza17, line 139: Every payload here hardcodes conversation_id: "conv-1", so the two real bugs above go untested: conversation_id=None auto-create (assert a conversation is created and done.conversation_id is non-null) and user_id attribution on the saved message. DoD requires tests for new features; these two would stop the gaps in this review from regressing past CI.

backend/src/taskorbit/api/routes/conversations.py
```diff
@@ -135,6 +137,95 @@ async def process_conversation(
...
+        if saved is None:
+            logger.error(
+                "sse_failed_to_save_assistant_message",
+                conversation_id=request.conversation_id,
+            )
+
+    yield f"data: {json.dumps({'type': 'done', 'intent': meta.selected_intent, 'status': meta.status, 'selected_agent': meta.selected_agent, 'slots': meta.extracted_slots, 'missing_slots': meta.missing_slots, 'conversation_id': meta.conversation_id, 'locked_intent_name': meta.locked_intent_name, 'next_active_tool_id': meta.next_active_tool_id, 'tool_invoked': meta.tool_invoked.model_dump() if meta.tool_invoked else None})}\n\n"
+
+
+@router.post("/stream")
```
Asadraza17, line 211: #57's DoD requires the System Architecture + SBOM docs to be updated, but the diff touches no docs while adding this public POST /v1/conversations/stream endpoint. Please add the endpoint + SSE event schema to the architecture doc and state that no new dependencies were added, so SBOM is unchanged.

Asadraza17 (changes requested):
Comments added.

Asadraza17 (approved):
Re-reviewed: all 9 points addressed cleanly. 5 new tests cover the gaps and pass locally. Nice work, approving

PR #140, opened 2026-06-11
Feature/custom agent routing

### LINKED ISSUE #129: Custom Agent Routing
## User story
1. As a user
2. I want the routing mechanism to support both automatic and manual forwarding to self defined Custom Agents
3. So that specialized user intents can be handled by dedicated agents rather than only built-in Agents

## Acceptance criteria
* The existing routing system is extended to recognize and process saved Custom Agents as valid routing targets
* The routing mechanism should support user-initiated transfers (e.g. via the UI by targeting a specific Agent ID or name), allowing users to manually dictate the handoff instead of relying solely on automatic AI intent recognition
* Conversation context and session metadata are passed over to the target Custom Agent upon routing execution
* It's build directly up on the Agent Routing Foundation (#8) without breaking compatibility for existing routing

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

qiblatainf:
## Summary

This PR implements UI-initiated manual agent routing, allowing users to explicitly transfer a conversation to a specific agent (built-in or custom) without relying on intent detection. It also extends the agent registry to resolve user-defined agents from the database, and hardens the implementation against security, recursion, and reliability issues.

---

### Backend — New Types (`types.py`)
- Added `ManualTransferRequest` model with `target_agent_id` (UUID of a saved `AgentConfiguration`) and `target_agent_name` (human-readable fallback). A `model_validator` enforces that at least one field is non-empty — an empty `ManualTransferRequest {}` is rejected at deserialization.
- Added `ToolType.EXTERNAL_API` enum value.

### Backend — Orchestration (`orchestration/__init__.py`)
- `process_message` now accepts optional `db: AsyncSession` and `user_id: int | None` parameters, enabling DB-backed agent lookups scoped to the authenticated user.
- New **step 0a**: if `request.manual_transfer` is set, the pipeline short-circuits intent detection and delegates to `_handle_manual_transfer`, which:
  1. Resolves the target by ID (user copy → default template) or by name (scoped to `user_id`).
  2. Loads and validates the target `AgentConfig` (invalid configs return a structured error response instead of raising).
  3. Re-enters `process_message` with `manual_transfer=None` cleared on the updated request, preventing infinite recursion.
- `user_id` is threaded through all recursive re-entries so the scope is never silently dropped.

### Backend — Agent Registry (`agents/__init__.py`)
- Added `CustomAgent` class: a thin `BaseAgent` wrapper for DB-loaded agents whose behaviour is driven entirely by their config.
- `AgentRegistry.create_by_name` is now `async` and accepts `db` and `user_id`. Resolution order: (1) built-in registry match, (2) DB lookup scoped to `user_id`, (3) default agent fallback. Corrupt `record.config` dicts are caught and fall back gracefully instead of crashing the pipeline.
- Added `AgentRegistry.known_agent_names()` returning the set of built-in `agent_name` strings.
- Added `AgentRegistry.create_custom()` for constructing a `CustomAgent` from an already-resolved config.

### Backend — Database (`database/crud.py`)
- `get_agent_configuration_by_id(db, agent_id)` — look up a saved config by primary key.
- `get_agent_configuration_by_name(db, name, user_id=None)` — look up by name, optionally scoped to a user. Used in both intent-driven DB fallback and manual transfers.

### Backend — Agent Transfer Tool (`tools/agent_transfer.py`)
- `AgentTransferTool.__init__` now accepts an optional `db` session.
- New `_is_valid_target(target_agent_id)` async method checks built-in names first, then falls back to DB lookup when `db` is available — so LLM-triggered transfers also work for custom agents.

### Backend — Route (`api/routes/conversations.py`)
- Passes the authenticated `user_id` to `orchestrator.process_message` so DB lookups are always scoped to the correct user.

### Backend — Voice Path (`livekit_agent/llm.py`)
- `AsyncSessionLocal` is now opened **before** `process_message` in `llm_node` so that manual transfers and custom-agent DB lookups are also available in voice calls. The existing persistence block uses a separate independent session as before.

### Frontend — `InCallControls.tsx`
- Added `@mention` routing syntax in the text input: typing `@` opens an agent-selection popover listing all available agents (built-in + custom). Selecting one locks the next message to that agent via `manual_transfer`.
- The routing target is cleared after each send and after voice recording starts.

### Frontend — `ConversationalChat.tsx`
- Wires the selected routing target from `InCallControls` into the `sendMessage` payload as `manual_transfer: { target_agent_id, target_agent_name }`.
- Appends an inline `↔ transferred to [Agent]` transcript marker when the backend responds with a different `selected_agent`.

### Frontend — `conversationApi.ts`
- `sendMessage` now accepts an optional `manualTransfer` argument and includes it in the POST body.

---

## Tests Added

| File | Coverage |
|---|---|
| `test_custom_agent_routing.py` | `ManualTransferRequest` validation (5 cases), `AgentRegistry.create_by_name` DB fallback with user_id scoping, corrupt-config fallback, `_handle_manual_transfer` recursion guard, name-lookup scoping, not-found and invalid-config error responses |
| `test_agent_transfer.py` | `AgentTransferTool._is_valid_target` for built-in and custom agents |
| `test_agent_runtime.py` | `CustomAgent` construction via `create_custom` and `create_by_name` DB path |
| `test_orchestration.py` | `_handle_manual_transfer` by-ID and by-name paths, recursion guard, error responses |

---

## Test Plan

- [x] Transfer to a custom DB agent by name — confirm the new agent's persona is applied from the next turn onward.
- [x] Send `manual_transfer: {}` (both fields empty) — confirm the API returns HTTP 422.
- [x] Delete a custom agent and attempt to transfer to it by name — confirm a structured error response (not a 500).
- [x] Run voice call and issue a manual transfer via the text input during the call — confirm it works (voice path now has DB access).
- [x] Run `poetry run pytest` — all tests pass.

### DIFF
--- a/backend/src/taskorbit/database/crud.py
+++ b/backend/src/taskorbit/database/crud.py
@@ -731,13 +731,21 @@ async def copy_on_write_user_agent(
 
 
 async def get_agent_configuration_by_id(
-    db: AsyncSession, agent_id: str
+    db: AsyncSession, agent_id: str, user_id: int | None = None
 ) -> AgentConfiguration | None:
-    """Look up a single AgentConfiguration by its primary key (async)."""
+    """Look up a single AgentConfiguration by its primary key (async).
+
+    When user_id is provided, only returns records owned by that user OR global
+    template records (user_id IS NULL). Returns None for both missing rows and
+    rows owned by another user — callers cannot distinguish the two cases.
+    """
     try:
-        result = await db.execute(
-            select(AgentConfiguration).where(AgentConfiguration.id == agent_id)
-        )
+        query = select(AgentConfiguration).where(AgentConfiguration.id == agent_id)
+        if user_id is not None:
+            query = query.where(
+                (AgentConfiguration.user_id == user_id) | (AgentConfiguration.user_id.is_(None))
+            )
+        result = await db.execute(query)
         return result.scalar_one_or_none()
     except SQLAlchemyError as e:
         logger.error("get_agent_configuration_by_id_failed", agent_id=agent_id, error=str(e))
@@ -747,11 +755,20 @@ async def get_agent_configuration_by_id(
 async def get_agent_configuration_by_name(
     db: AsyncSession, name: str, user_id: int | None = None
 ) -> AgentConfiguration | None:
-    """Look up a single AgentConfiguration by name, optionally scoped to a user."""
+    """Look up a single AgentConfiguration by name, scoped to a user.
+
+    When user_id is provided, matches records owned by that user only.
+    When user_id is None (e.g. voice path before auth lands), fails closed:
+    only global template records (user_id IS NULL) are returned so the voice
+    path cannot resolve one user's custom agent into another user's session.
+    """
     try:
         query = select(AgentConfiguration).where(AgentConfiguration.name == name)
         if user_id is not None:
             query = query.where(AgentConfiguration.user_id == user_id)
+        else:
+            # Fail-closed: unauthenticated callers see global templates only.
+            query = query.where(AgentConfiguration.user_id.is_(None))
         result = await db.execute(query)
         return result.scalar_one_or_none()
     except SQLAlchemyError as e:
--- a/backend/src/taskorbit/database/crud.py
+++ b/backend/src/taskorbit/database/crud.py
@@ -730,6 +730,37 @@ async def copy_on_write_user_agent(
         return None
 
 
+async def get_agent_configuration_by_id(
+    db: AsyncSession, agent_id: str
+) -> AgentConfiguration | None:
+    """Look up a single AgentConfiguration by its primary key (async)."""
+    try:
+        result = await db.execute(
+            select(AgentConfiguration).where(AgentConfiguration.id == agent_id)
+        )
+        return result.scalar_one_or_none()
+    except SQLAlchemyError as e:
+        logger.error("get_agent_configuration_by_id_failed", agent_id=agent_id, error=str(e))
+        return None
+
+
+async def get_agent_configuration_by_name(
+    db: AsyncSession, name: str, user_id: int | None = None
+) -> AgentConfiguration | None:
+    """Look up a single AgentConfiguration by name, optionally scoped to a user."""
+    try:
+        query = select(AgentConfiguration).where(AgentConfiguration.name == name)
+        if user_id is not None:
+            query = query.where(AgentConfiguration.user_id == user_id)
+        result = await db.execute(query)
+        return result.scalar_one_or_none()
+    except SQLAlchemyError as e:
+        logger.error(
+            "get_agent_configuration_by_name_failed", name=name, user_id=user_id, error=str(e)
+        )
+        return None
+
+
 async def list_user_agents_merged(db: AsyncSession, user_id: int) -> list[dict]:
     """Return all agents for a user — their customised copies AND all templates.
 
--- a/backend/src/taskorbit/livekit_agent/llm.py
+++ b/backend/src/taskorbit/livekit_agent/llm.py
@@ -239,7 +239,20 @@ class OrchestratorAgent(Agent):
             messages=messages,
             current_intent_name=self._locked_intent_name,
         )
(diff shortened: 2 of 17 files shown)


backend/src/taskorbit/orchestration/__init__.py
Asadraza17: AgentTransferTool() is constructed here with no db, so _db is always None and the custom-agent branch in _is_valid_target is dead in production (the passing test hand-injects a db). Custom agents can't be reached via the automatic transfer path. Suggest threading db through _dispatch_tool and constructing tool_cls(db=db) for AGENT_TRANSFER.

Asadraza17: This slug-normalization is for built-in names, but it mangles a custom-agent UUID (replace("-","_") corrupts the hyphens), so the auto path can never match a custom id even once the db is wired. Skip normalization for values that look like custom ids.

backend/src/taskorbit/database/crud.py
```diff
@@ -730,6 +730,37 @@ async def copy_on_write_user_agent(
         return None
 
 
+async def get_agent_configuration_by_id(
```
Asadraza17, line 733: This lookup takes no user_id, but _handle_manual_transfer feeds it the client-supplied target_agent_id and runs the LLM under whatever it loads. by-name is scoped; by-id should be too (and by-id takes precedence). Suggest scoping with (user_id == user_id) | (user_id.is_(None)) and returning a generic not-found to avoid an existence oracle.

backend/src/taskorbit/livekit_agent/llm.py
```diff
@@ -239,7 +239,20 @@ async def llm_node(  # type: ignore[override]
             messages=messages,
             current_intent_name=self._locked_intent_name,
         )
-        response = await self._orchestrator.process_message(request)
+
+        # Open the DB session before process_message so that manual transfers
+        # and custom-agent DB lookups work in the voice path too.
+        try:
+            async with AsyncSessionLocal() as db:
+                response = await self._orchestrator.process_message(request, db=db)
```
Asadraza17: Voice path passes no user_id here, so by-name custom-agent resolution is unscoped (matches across users). Needs user_id threaded into OrchestratorAgent, or have the by-name lookup fail closed when user_id is None.

frontend/src/components/ConversationalChat.tsx
```diff
@@ -338,9 +372,38 @@ export function ConversationalChat() {
   );
 
   const handleVoiceAgentRouted = useCallback((agentName: string) => {
+    if (manualRoutingLockRef.current) return;
     setRoutedAgent(agentName);
   }, []);
 
+  // Fires the moment the user picks an agent from the @route dropdown — no send needed.
+  const handleRoutingTargetChange = useCallback(
```
Asadraza17, line 391: This fires the transfer side effects (marker, TTS, swap) on select, and handleSendText (line 269) fires them again on send, so a normal pick-then-send double-transfers. Suggest one owner for the swap + marker.

frontend/src/components/ConversationalChat.tsx
```diff
@@ -243,17 +249,51 @@ export function ConversationalChat() {
...
             return;
           }
 
+          // Manual transfer: swap active agent config and show badge/marker.
+          if (manualTransfer?.target_agent_id && response.status !== "error") {
+            // The badge appends " Agent", so strip it from the stored name to avoid "Sales Agent Agent".
+            const badgeName = manualTransfer.target_agent_name.replace(/\s+[Aa]gent$/i, "").trim();
+            manualRoutingLockRef.current = true;
+            setRoutedAgent(badgeName);
+            call.appendAssistantTurn(`[Transferred to ${manualTransfer.target_agent_name}]`);
```
Asadraza17: If the target isn't found (stale list / deleted / backend rejects) this still appends "[Transferred to X]" and flips the badge, so the UI claims a transfer that didn't happen. Add a failure marker + revert on no-match.

Asadraza17 (changes requested):
Comments added.

Asadraza17 (approved):
Approved!

Issue #99, opened 2026-05-27
Agent Config not persistant through text and voice prompts


Perimora:
**Describe the bug**
Text prompts use the expected agent configuration loaded via User FE. When executed via voice prompts the agent defaults to old config personas. 

**To Reproduce**
1. Checkout main
2. Create new agent config
3. Start session with new agent
4. Ask specific question regarding new config
5. Guardrails prevent agent from answering

**Expected behavior**
Voice and text prompts lead to the same expected behavior in compliance with dynamic configs

**Desktop:**
 - OS: Win11
 - Browser Opera GX
 - Version 131.0.5877.84

PR #141, opened 2026-06-12
feat: interchangeable STT/TTS provider selection (#135)

### LINKED ISSUE #135: Interchangeable Model Selection
## User story
1. As a user
2. I want to interchange STT and TTS providers
3. So that I can identify the optimal speech pipeline for my use case

## Acceptance criteria
* The user can select available STT providers independently of available TTS providers
* Providers that support both STT and TTS can be used for either capability or for both
* The selected STT and TTS providers are applied consistently during pipeline execution
* The provider selection can be changed without modifying the underlying pipeline configuration manually

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

Asadraza17:
Closes #135

Agents can now use Deepgram or ElevenLabs for STT and TTS independently, selected per agent in the config UI. The selection travels config UI -> saved agent -> room metadata -> worker, where a provider-dispatch factory builds the actual plugins. Switching providers auto-resets the model to that provider's default (same guard as #99). Zero new dependencies, zero new env vars.

All 4 combinations live-verified on the local stack (2026-06-12), including ElevenLabs Scribe realtime STT on a free-tier key. Every session logs its resolved selection (stt_selected / tts_selected) for easy review.

**Behavior change to be aware of:** the agent config now wins over env for voice_id and models (env stays the fallback). Previously env always won, which made the config voice field a non-functional placeholder. Default seed agents still carry the Rachel voice, which 402s on free ElevenLabs keys; swap to a free voice proposed separately.

Tests: 22 new (provider matrix, model pinning, voice precedence, worker metadata round-trip). 
Docs: new provider selection guide.

### DIFF
--- a/Documentation/stt-tts-providers.md
+++ b/Documentation/stt-tts-providers.md
@@ -1,0 +1,113 @@
+# Interchangeable STT/TTS Providers
+
+> **Scope.** This guide explains how to choose the speech-to-text and
+> text-to-speech provider per agent (ticket #135). Selection happens in
+> the Agent Config UI; no code or environment change is needed to switch.
+>
+> The dispatch logic lives in
+> `backend/src/taskorbit/livekit_agent/session.py` (`_build_stt` /
+> `_build_tts`).
+
+---
+
+## What this is
+
+Before #135 the voice pipeline was hardwired: Deepgram for STT, ElevenLabs
+for TTS, both configured purely from environment variables. The provider
+dropdowns in the Agent Config UI existed but had a single option, and the
+values were discarded on the way to the voice worker.
+
+Now both vendors are dual-capability and every combination works:
+
+| | TTS: ElevenLabs | TTS: Deepgram (Aura) |
+|---|---|---|
+| **STT: Deepgram** | historical default | works |
+| **STT: ElevenLabs (Scribe)** | works | works |
+
+The selection travels: Agent Config UI -> saved agent -> LiveKit token
+metadata -> voice worker -> plugin construction. Changing providers is a
+dropdown choice, saved with the agent (AC4: no manual pipeline edits).
+
+## Selecting providers in the UI
+
+Pipeline section of the Agent Config page:
+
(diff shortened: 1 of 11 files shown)


qiblatainf (commented):
Fix before merge:                                                                                                                                                                                                                                                                                                                                                                                                         
-   The cross-provider model detection in session.py is asymmetric — a Deepgram model like nova-3 passed to ElevenLabs TTS slips through undetected and fails silently at the API. The fix is a shared helper so all  four branches (Deepgram STT, ElevenLabs STT, Deepgram TTS, ElevenLabs TTS) use the same detection pattern.                                                                                                         
   
Worth fixing:                                                                                                                                                                                                        
  1. is vs == for enum comparison (session.py) — Pydantic can return new enum instances, is can fail. Use ==.                                                                                                        
  2. Add a warning log assertion in the fallback tests — right now they verify the fallback model is used but don't check that a warning was actually logged, which is the only signal for misconfiguration in production.

Asadraza17:
Thanks for the review @qiblatainf, all three addressed in the latest commit. 

1. Pulled the cross-provider model check into one _resolve_model helper that all four branches (Deepgram STT, ElevenLabs STT, Deepgram TTS, ElevenLabs TTS) now route through, each with its own predicate. One note on the example: nova-3 to ElevenLabs TTS was actually already caught and warned. The branch that genuinely let things slip through silently was Deepgram STT, it only blocked scribe_ names and passed everything else to the API. That's the one this fixes: it now allowlists nova- models and falls back with a warning on anything else (this does tighten Deepgram STT to the Nova family, so a non-Nova Deepgram model typed into the free-text field would fall back to nova-3 with a warning, flagging in case we want to widen it later).

2. Changed both enum comparisons from is to ==.

3. The four fallback tests now assert the warning actually fired, plus a valid-model test that asserts it stays silent so the check isn't vacuous plus direct tests for the helper.

qiblatainf (approved):

Issue #71, opened 2026-05-15
Conditional Workflow Engine


Caerum5:
## User story
1. As a user
2. I want the agent to follow a logical, multi-step workflow when handling complex requests
3. So that my request is processed reliably and accurately without losing context or becoming confused.

## Acceptance criteria
* At least two independent conditional workflows can be executed in the extended runtime environment
* The engine supports branching into different execution paths
  * If condition A is met, workflow block X is executed
  * Otherwise, workflow block Y is executed
* The engine exclusively uses custom agent definitions. No hardcoded agents are used
* For each agent, the user can specify its workflow dependencies (e.g. prerequisite agent definitions) in the frontend
* The user can define the allowed handoffs between agent definitions in the frontend
* When an agent is instantiated, the backend resolves the configured workflow dependencies and builds the execution graph
* If workflow dependencies exist, the agent informs the user about the planned workflow execution and requests confirmation before executing the prerequisite workflow steps
* If the user confirms, the workflow engine executes the required workflow steps before continuing with the originally instantiated agent
* If a workflow step cannot be executed due to an unsatisfied dependency, a predefined fallback mechanism prevents workflow deadlocks
* The workflow builds upon the Agent Routing Foundation (#8) and evaluates the conditional workflow logic after the routing decision has been completed

## Definition of done (DoD)
* General coding standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

PR #142, opened 2026-06-15
Feature/Conditional workflow engine

shikharthakur2404:
_**### Summary**_
Implements the Conditional Workflow Engine (#71): users define workflow prerequisites and allowed handoffs per saved agent in Agent Config; the backend orchestrator enforces prerequisite confirmation, executes the dependency agent after Proceed, and gates handoffs against the allow-list.

Builds on Agent Routing (#8): intent routing runs first; workflow and handoff rules apply on subsequent turns.

**### User story**
As a user, I want the agent to follow a logical, multi-step workflow when handling complex requests, so that my request is processed reliably and accurately without losing context or becoming confused.

**### What’s in this PR**

Frontend
Workflow section on Agent Config: prerequisite steps + allowed handoffs (saved-agent dropdown)
Cycle detection (self-dependency + circular deps) with inline errors
Update / Save as new disabled while workflow validation fails
Chat: Prerequisite step required card (Proceed / Cancel) on text path
Agent swap after prerequisite Proceed; completed_workflow_steps tracked per session

Backend
workflow_dependencies / allowed_handoffs on AgentConfig
Turn-1 entry-agent lock (configured agent owns first turn)
Workflow confirmation before running prerequisites
Prerequisite execution after Proceed (no re-prompt loop on tech questions)
Handoff allow-list with config-id → registry-name mapping
Explicit handoff_blocked response status
Tests & schema
backend/tests/test_workflow_engine.py (7 tests)
backend/tests/test_types.py — workflow fields
Schema: workflow_dependencies, allowed_handoffs

`Delivered by Shikhar & Endri `


backend/src/taskorbit/agents/__init__.py
```diff
@@ -167,6 +167,15 @@ class AgentRegistry:
 
     _DEFAULT: type[BaseAgent] = SalesAgent
 
+    @classmethod
+    def get_agent_name_for_id(cls, config_id: str) -> str:
+        """Map a logical config.id (e.g. from allowed_handoffs) to a registry agent_name."""
+        agent_id = config_id.lower()
+        for keywords, agent_cls in cls._REGISTRY:
+            if any(kw in agent_id for kw in keywords):
```
djv03, line 198: AC3- partially fulfil 
Since AC3 mentioned routing through custom agents, not hardcoded agent from the registry.

djv03 (commented):
PR delivers AC2, AC4, AC5, AC7 and AC10. These are well implemented and tested.

However, these ACs have blocking gaps:

AC3:  Hardcoded agents are still in the resolution path.

The new `get_agent_name_for_id` method added by this PR routes custom workflow dependency IDs through `_REGISTRY` and falls back silently to `SalesAgent._DEFAULT` when no keyword matches. 
A user-defined dep like `"onboarding-flow"` will silently execute as a Sales agent with no error. Workflow dep resolution should look up IDs from `AgentConfig` directly, not from the hardcoded registry.


Additional issues to address before merge:
- `handoff_blocked` is returned by the backend but missing from the frontend `ConversationResponse` status union in `conversationApi.ts`
- AC9 fallback only covers user rejection; technical failures (unresolvable dep ID, agent throws) have no deadlock guard
- The agent card swap after confirmation silently no-ops for custom dep IDs that don't match any backend record

Happy to merge once changes are adressed

1Endri:
Thanks for the thorough review. All four points addressed:

- AC3 — replaced the silent SalesAgent fallback with a GenericAgent that preserves 
  the custom ID; workflow deps now resolve from the DB config, not the hardcoded 
  registry.
- AC9 — added deadlock guards for unresolvable dep IDs / technical failures 
  (resets to entry agent + returns handoff_blocked).
- handoff_blocked — now in the frontend ConversationResponse status union, and the 
  refusal message displays correctly.
- Agent card swap — now works for custom dep IDs since the backend returns the real 
  ID (the frontend lookup was correct, it was just being fed "sales" before).

All 428 backend tests passing, frontend type-check + lint clean. Ready for re-review.

djv03 (approved):
All ACs are satisfying as per backlog description

Approving anf merging PR

PR #148, opened 2026-06-16
feat(#87): provider-aware model and voice dropdowns for STT/LLM/TTS

### LINKED ISSUE #87: Provider Model Dropdown
## User story
1. As a user
2. I want to select models from a predefined dropdown menu based on the selected provider
3. So that I don't have to manually type exact model names, preventing typos and runtime configuration errors

## Acceptance criteria
* The "Model" text input field is replaced by a dropdown selection menu
* Changing the provider (e.g. switching from OpenAI to Gemini or Deepgram) instantly updates the available options in the model dropdown to reflect only that specific provider's supported models
* The backend must be accept the selected model dynamically from the frontend configuration/API request instead of fallback-loading a single hardcoded model from the .env file
* The dropdown contains standard models for your initial providers 
* If no model is selected or a legacy configuration is loaded, the system can still use the current .env value (like gpt-4o-mini) as a default fallback so that nothing breaks

## Definition of done (DoD)

## DoD general criteria
* All tests are passing

K-Vineesh:
Closes #87.

## Summary
Replaces the free-text model inputs in the agent config UI with provider-aware dropdowns for all three pipeline stages (STT, LLM, TTS), plus a voice selector showing readable names instead of raw ElevenLabs IDs. Per the PO's scope update on this ticket, this covers all three stages rather than just LLM.

## Changes

**Frontend:**
- New `frontend/src/lib/pipelineOptions.ts` — single source of truth for selectable models and voices. Exports `LLM_MODELS`, `STT_MODELS`, `TTS_MODELS`, `TTS_VOICES`, their defaults, and a `withCurrent()` helper. Adding a new model going forward means editing this one file.
- `PipelineSection.tsx`: all three Model fields converted from text `<Input>` to `<Select>` dropdowns, scoped to the current provider's models. The Voice ID text input is replaced with a `<Select>` showing readable names (Roger / Jessica / Chris / Eric — free-tier ElevenLabs voices) while storing the underlying voice ID.
- Switching provider on any of the three stages resets the model to that provider's default (the LLM stage already did this from #99; STT and TTS now match).
- AC5 (legacy/no-selection fallback): if a saved config holds a model or voice ID that isn't in the current list, the dropdown still shows that value as a selectable extra option instead of blanking it, so old configs aren't silently changed.

**Backend (real bug found during testing, not just dropdown wiring):**
- `build_agent_session()` in `session.py` accepted an `agent_config` parameter but never used it — the ElevenLabs TTS and Deepgram STT plugins were built purely from `.env` settings regardless of what was selected per-agent. This meant a user's dropdown selection was parsed correctly from the LiveKit JWT metadata but silently discarded before reaching the actual audio plugins.
- Fixed by threading `agent_config.stt.model`, `agent_config.tts.voice_id`, and `agent_config.tts.model` into the plugin construction, falling back to env defaults when the config or a field is absent (AC5).
- `worker.py` now passes the parsed `agent_config` into `build_agent_session()`.
- Added two regression tests in `test_livekit_agent_tts.py`: one confirming the metadata override path, one confirming the env-fallback path.
- Note: the LLM model was already correctly wired (verified separately) — `OrchestratorAgent.llm_node` reads `agent_config.llm.model` directly and was never affected by this bug; only the audio plugins were.

## Testing
- `npx tsc --noEmit` clean.
- `poetry run pytest tests/test_livekit_agent_tts.py -v` passing.
- Manually verified: selecting a voice/model in the UI, saving, and starting a voice session now actually uses the selected voice and STT/TTS model rather than the env default.
- Free-tier ElevenLabs voices used (Roger, Jessica, Chris, Eric) since the project account is on the free tier.

### DIFF
--- a/frontend/src/components/agent-config/PipelineSection.tsx
+++ b/frontend/src/components/agent-config/PipelineSection.tsx
@@ -235,32 +235,35 @@ export function PipelineSection({ value, onChange }: Props) {
                 </SelectContent>
               </Select>
             </Field>
-            <Field>
-              <FieldLabel htmlFor={idTtsVoice}>Voice</FieldLabel>
-              <Select
-                value={value.tts.voice_id}
-                onValueChange={(v) => onChange({ ...value, tts: { ...value.tts, voice_id: v } })}
-              >
-                <SelectTrigger id={idTtsVoice}>
-                  <SelectValue />
-                </SelectTrigger>
-                <SelectContent>
-                  <SelectGroup>
-                    {TTS_VOICES.map((voice) => (
-                      <SelectItem key={voice.id} value={voice.id}>
-                        {voice.name}
-                      </SelectItem>
-                    ))}
-                    {/* Legacy / unlisted voice id — show the raw id so it round-trips (AC5). */}
-                    {!voiceIsKnown && value.tts.voice_id ? (
-                      <SelectItem value={value.tts.voice_id} className="font-mono text-sm">
-                        {value.tts.voice_id}
-                      </SelectItem>
-                    ) : null}
-                  </SelectGroup>
-                </SelectContent>
-              </Select>
-            </Field>
+            {/* TTS_VOICES are ElevenLabs-only; Deepgram encodes voice in the model name. */}
+            {value.tts.provider === "elevenlabs" && (
+              <Field>
+                <FieldLabel htmlFor={idTtsVoice}>Voice</FieldLabel>
+                <Select
+                  value={value.tts.voice_id}
+                  onValueChange={(v) => onChange({ ...value, tts: { ...value.tts, voice_id: v } })}
+                >
+                  <SelectTrigger id={idTtsVoice}>
+                    <SelectValue />
+                  </SelectTrigger>
+                  <SelectContent>
+                    <SelectGroup>
+                      {TTS_VOICES.map((voice) => (
+                        <SelectItem key={voice.id} value={voice.id}>
+                          {voice.name}
+                        </SelectItem>
+                      ))}
+                      {/* Legacy / unlisted voice id — show the raw id so it round-trips (AC5). */}
+                      {!voiceIsKnown && value.tts.voice_id ? (
+                        <SelectItem value={value.tts.voice_id} className="font-mono text-sm">
+                          {value.tts.voice_id}
+                        </SelectItem>
+                      ) : null}
+                    </SelectGroup>
+                  </SelectContent>
+                </Select>
+              </Field>
+            )}
           </FieldGroup>
         </FieldSet>
       </CardContent>
--- a/frontend/src/components/agent-config/PipelineSection.tsx
+++ b/frontend/src/components/agent-config/PipelineSection.tsx
@@ -4,7 +4,6 @@ import type { LucideIcon } from "lucide-react";
 
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from "@/components/ui/card";
 import { Field, FieldGroup, FieldLabel, FieldLegend, FieldSet } from "@/components/ui/field";
-import { Input } from "@/components/ui/input";
 import {
   Select,
   SelectContent,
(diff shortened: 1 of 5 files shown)


frontend/src/lib/pipelineOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+import type { LlmProvider, SttProvider, TtsProvider } from "@/types/agentConfig";
+
+export const LLM_MODELS: Record<LlmProvider, string[]> = {
+  openai: ["gpt-4o-mini", "gpt-4o"],
+  gemini: ["gemini-2.5-flash", "gemini-2.5-pro"],
+};
+
+// First model of each provider's list is the default applied on provider switch.
+export const LLM_MODEL_DEFAULTS: Record<LlmProvider, string> = {
+  openai: LLM_MODELS.openai[0],
```
abdulmoeez1225, line 12: Default is coupled to array index 0 — if someone reorders `LLM_MODELS.openai`, the provider-switch default silently changes with no indication at this call site. Either hardcode the default string explicitly (`openai: "gpt-4o-mini"`) or add a comment warning that index 0 = default.

frontend/src/lib/pipelineOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+  gemini: LLM_MODELS.gemini[0],
+};
+
+export const STT_MODELS: Record<SttProvider, string[]> = {
+  deepgram: ["nova-3", "nova-2"],
+  elevenlabs: ["scribe_v2_realtime"],
+};
+
+export const STT_MODEL_DEFAULT: string = STT_MODELS.deepgram[0];
```
abdulmoeez1225: `STT_MODEL_DEFAULT` is exported but never imported in `PipelineSection.tsx` or anywhere else — dead export. Remove it, or replace with `STT_MODEL_DEFAULTS: Record<SttProvider, string>` and wire it into the STT provider-switch `onValueChange` handler (line 88 of `PipelineSection.tsx`).

frontend/src/lib/pipelineOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const TTS_MODELS: Record<TtsProvider, string[]> = {
+  elevenlabs: ["eleven_multilingual_v2", "eleven_flash_v2_5"],
+  deepgram: ["aura-2-andromeda-en", "aura-2-asteria-en", "aura-2-orion-en"],
+};
+
+export const TTS_MODEL_DEFAULT: string = TTS_MODELS.elevenlabs[0];
```
abdulmoeez1225: `TTS_MODEL_DEFAULT` is exported but never imported anywhere — dead export. Same fix as `STT_MODEL_DEFAULT`: remove or replace with a keyed Record and wire into the TTS provider-switch handler.

frontend/src/lib/pipelineOptions.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Returns `options` with `current` appended when it holds a value that isn't
+ * already listed. Keeps legacy / env-default values selectable so loading an
+ * older config never drops the user's stored choice from the dropdown (AC5).
+ */
+export function withCurrent(options: string[], current: string): string[] {
+  if (current && !options.includes(current)) {
+    return [...options, current];
+  }
+  return options;
+}
```
abdulmoeez1225: `options` can be `undefined` at runtime. `STT_MODELS[value.stt.provider]` returns `undefined` if a saved config in localStorage carries a provider string that isn't a key in the Record (e.g. a legacy or renamed provider). `options.includes(current)` then throws `TypeError: Cannot read properties of undefined`. Fix the signature and add a fallback:

```ts
export function withCurrent(options: string[] | undefined, current: string): string[] {
  const base = options ?? [];
  if (current && !base.includes(current)) return [...base, current];
  return base;
}

---

File: frontend/src/components/agent-config/PipelineSection.tsx [hunk 5]
```diff
@@ -189,30 +217,50 @@ export function PipelineSection({ value, onChange }: Props) {
...
-                    onChange({ ...value, tts: { ...value.tts, voice_id: e.target.value } })
-                  }
-                  placeholder="JBFqnCBsd6RMkjVDRZzb"
-                  className="font-mono text-sm"
-                />
-              </Field>
-            )}
           </FieldGroup>
         </FieldSet>
       </CardContent>
```
abdulmoeez1225, line 269: (the `<Field>` opening the Voice dropdown)
Bug: This Voice dropdown must be wrapped in {value.tts.provider === "elevenlabs" && (...)}. TTS_VOICES contains only ElevenLabs voice IDs. Currently, when TTS provider is deepgram the dropdown still renders and a user can save an ElevenLabs voice_id against a Deepgram config. session.py:59 then forwards that ID into elevenlabs.TTS() even though Deepgram is active — a silent misconfiguration that causes a broken or fallen-back voice call.

K-Vineesh:
Hii Abdul, Thank You for the Review , Please do rereview for the updated changes, 
Addressed all 5 points plus the missing test in <hash>:
- Voice dropdown now guarded to elevenlabs only (blocker)
- withCurrent handles undefined options
- removed the dead STT_MODEL_DEFAULT/TTS_MODEL_DEFAULT exports
- added the index-0-is-default comment
- condensed the header comment
- added the partial-config fallback test

abdulmoeez1225 (approved):

PR #157, opened 2026-06-19
Feature/71 conditional workflow engine

### LINKED ISSUE #71: Conditional Workflow Engine

shikharthakur2404:
Closes #71

## Summary

- Implement the **conditional workflow engine**: static prerequisite steps, **`workflow_rules`** branching by intent/routed agent, and **transitive dependency expansion** (if A → B and B → C, runtime order is C → B → A without listing C on A).
- Fix orchestration so prerequisite steps use the **correct agent persona** in the LLM, **`continue`** after Proceed does not hit intent clarification, and the **entry agent** is resumed after all prereqs complete.
- Keep **voice and UI in sync** via LiveKit `workflow_state` when the user clicks **Proceed** during an active voice session.
- Fix **Save as new** agent persistence: `POST /v1/user-agents` creates independent rows; **`agent_id`** in config blobs; prerequisite dropdown loads from **My agents**; UI title stays on entry agent name (not intent registry names like `general_inquiry`).
- Add **tests** for workflow rules, transitive enrich/expand, full C→B→A chain, and user-agents API regression.
- Add **documentation**: transitive workflow guide, demo scripts, QA handoff, GitHub wiki source.

## What to demo

**Transitive chain (voice):** configure `agent-c` → `agent-b` → `agent-a` (or `identity-verification` → `sales-qualification` → `customer-entry`), start session on entry agent, speak *I want to buy something*, **Proceed** through C then B, say *continue* each step — expect **STEP-C → STEP-B → STEP-A** (or VERIFY → SALES → ENTRY).

See `Documentation/transitive-workflow-guide.md` and `Documentation/demo-2-transitive-chain.md`.

## Test plan

- [x] `cd backend && poetry run pytest tests/test_workflow_engine.py tests/test_workflow_rules.py tests/test_transitive_dependencies.py tests/test_crud_transitive.py tests/test_user_agents_routes.py -q`
- [x] Agent Config: **Save as new** creates three agents (C → B → A) without overwriting
- [x] Prerequisite dropdown on B lists C after C is saved (hard refresh)
- [x] Voice session on entry agent: first **Proceed** is **Step C**, not B
- [x] After each **Proceed**, *continue* returns correct persona prefix (not entry agent early)
- [x] Session title stays **Step A Entry** / **Customer Entry** throughout
- [x] UI **Proceed** during voice session keeps worker aligned (Demo 3)
- [x] `poetry run taskorbit-worker dev` + LiveKit keys in `.env`

## Notes for reviewers

- **@developer** — please run `Documentation/issue-71-endri-testing-handoff.md` on latest branch.
- Typed text input is legacy; manual demos are **voice-first** (mic → Send + Proceed). Text HTTP path remains for automated tests.
- Workflow confirmation card **left border** is intentional UI styling.


abdulmoeez1225 (commented):
### Overview

This PR adds a conditional workflow engine to TaskOrbit. The core capabilities are:

1. **Static prerequisite chains** — agents can declare prerequisite steps that must complete before they run
2. **Conditional branching via `workflow_rules`** — pick different prerequisite sets based on detected intent/routed agent
3. **Transitive dependency expansion** — if A→B and B→C, the engine automatically resolves C→B→A without A needing to list C
4. **Voice/UI sync** — UI Proceed events publish `workflow_state` on the LiveKit data channel so the voice worker stays aligned

---

### Strengths

- `workflow_rules.py` is clean, well-separated, and testable in isolation
- `expand_workflow_dependencies` handles cycles correctly via a `seen` set — no infinite loops possible
- `_effective_selected_agent()` normalises empty-string voice path artifacts before they propagate — good defensive coding
- `active_config = agent.config` fix is the right architectural fix and applied consistently across both `process_message` and `stream_message`
- `sync_workflow_state()` on `OrchestratorAgent` is a clean, safe mutation point

---

### Issues

#### ❌ 1. `dev_user_id = 1` hardcoded in `worker.py`

```python
# worker.py ~line 120
dev_user_id = 1   # ← violates project convention
```

*"Never hardcode `id=1` which is fragile against auto-increment variance."* Should use the same email-based lookup that `get_current_user_id()` uses:

```python
async with AsyncSessionLocal() as db:
    user = await get_user_by_email(db, "dev@taskorbit.local")
    dev_user_id = user.id if user else None
```

#### ❌ 2. `schemas/agent-task.schema.json` not updated

The new `workflow_rules` array (`WorkflowRuleWhen` + `WorkflowRule`) is not reflected in the schema. Any tooling that validates agent configs against this schema will silently reject configs using `workflow_rules`.

#### ⚠️ 3. Duplicated workflow resolution logic

The workflow dependency block (detect → resolve → expand → find missing) is copy-pasted verbatim in both `process_message` and `stream_message`. A shared private method would eliminate the duplication and the risk of the two diverging:

```python
def _resolve_missing_dependencies(
    self, request: ConversationRequest, intent: IntentResult
) -> tuple[list[str], list[str]]:
    """Returns (effective_dependencies, missing_dependencies)."""
    ...
```

#### ⚠️ 4. In-band imports inside hot path

`_KNOWN_INTENTS` and `workflow_rules` functions are imported inside `_is_executing_workflow_prerequisite()` and `_resolve_intent_after_clarification_gate()` on every call. These are module-level singletons — move them to the top of `orchestration/__init__.py`.

#### ⚠️ 5. `_current_routed_agent` type change (`str` → `str | None`)

Changed without a check for downstream consumers in `llm.py` that may do string operations on it. Please confirm no `.startswith()` / string concatenation uses remain on this field.

#### ⚠️ 6. `enrich_request_dependency_configs` uses `list.pop(0)` as a queue

```python
dep_id = to_check.pop(0)  # O(n) on every iteration
```

Use `collections.deque` with `popleft()` for O(1). Not a bug but degrades on deep chains.

---

### Missing Tests

- `_resolve_intent_after_clarification_gate` — no direct unit test; this is the "continue after Proceed" regression fix and the most likely to break silently
- `sync_workflow_state()` — no test verifies that a `workflow_state` data-channel message actually updates `_current_routed_agent` and `_completed_workflow_steps`
- `_effective_selected_agent("")` / `_effective_selected_agent("  ")` edge cases

---

### Summary

| Area | Status | Finding |
|---|---|---|
| Core engine | ✅ | Solid. Correct cycle handling, clean separation of concerns. |
| Orchestration | ⚠️ | Duplicated dependency resolution block across `process_message` / `stream_message` |
| In-band imports | ⚠️ | Move to module top level |
| `dev_user_id = 1` | ❌ | use email-based lookup |
| Schema | ❌ | `agent-task.schema.json` missing `workflow_rules` definition |
| Test gaps | ⚠️ | Missing tests for clarification gate and `sync_workflow_state` |

**Recommendation:** Fix `dev_user_id = 1` and update the schema before merge. The duplicated resolution block and in-band imports can be follow-up issues.

shikharthakur2404:
implemented

abdulmoeez1225:
## Overview

Implements a conditional workflow engine with static prerequisite steps, `workflow_rules` intent/agent-name branching, and transitive dependency expansion. Also fixes the entry-agent resume bug, voice/UI sync via LiveKit data channel, `Save as new` persistence, and adds the `POST /v1/user-agents` route. The core logic is sound and the architecture maps cleanly to the existing pipeline.

---

## Issues

### Critical

#### 1. `resolve_dependency_agent_config` does a full table scan (`crud.py:669–690`)

After the direct PK lookup misses, the fallback fetches every `AgentConfiguration` row for the user and scans in Python to match on the logical `agent_id` field inside the JSON blob:

```python
result = await db.execute(
    select(AgentConfiguration).where(
        (AgentConfiguration.user_id == user_id) | (AgentConfiguration.user_id.is_(None))
    )
)
for candidate in result.scalars().all():
    blob = candidate.config or {}
    if logical_id_from_stored_blob(blob) != logical_id:
        continue
```

This is O(n) memory and time as agents grow. Should push the filter to the DB via a JSON path expression (e.g. `config->>'agent_id'` or `config->>'id'`) or add a `logical_id` column. At minimum, add a `LIMIT 1` since we break on the first match.

---

### High

#### 2. `update_user_agent` uses naive `datetime.now()` (`crud.py:784`)

```python
agent.updated_at = datetime.now()
```

The codebase imports `from datetime import UTC, datetime` at the top of `crud.py` and every other timestamp uses `datetime.now(UTC)`. This one is naïve (no timezone) and will sort incorrectly if the DB ever compares it to UTC-aware values.

**Fix:** `agent.updated_at = datetime.now(UTC)`

---

#### 3. Massive DRY violation between `process_message` and `process_message_stream` (`orchestration/__init__.py`)

Both methods are ~500 lines each and share virtually identical logic for: end-call detection, intent detection, agent selection, workflow dependency resolution, handoff enforcement, slot extraction, system prompt building, and step tracking. Only the `yield`/`return` mechanics differ.

This is the most significant maintainability risk in the PR. A bug fix or logic change (e.g., in the clarification-gate fix at line 287/804) must be applied in two places and keeping them in sync will be error-prone. The two paths are already drifting — the non-streaming path logs `llm_call` latency after the await (line 545), the streaming path after the stream finishes (line 1060), but they share no abstraction.

**Suggestion:** Extract the shared pre-LLM pipeline (steps 0–4 + dependency resolution) into a private `_prepare_turn(request) -> _TurnContext` dataclass method, then let `process_message` and `process_message_stream` only differ in how they call the LLM and yield results.

---

#### 4. `_direct_dependencies_for_request` and `_resolve_missing_dependencies` compute deps from different sources (`orchestration/__init__.py:79–120`)

`_is_executing_workflow_prerequisite` calls `_direct_dependencies_for_request` which resolves rules from `_KNOWN_INTENTS[request.current_intent_name]`. But `_resolve_missing_dependencies` (called in the main pipeline) uses the freshly detected `intent` object. If the intent lock is broken (line 256–263), `intent.name != request.current_intent_name`, meaning the two helpers compute dependencies from **different intents** for the same turn. This can cause `_is_executing_workflow_prerequisite` to return `True` while the actual `missing_dependencies` list points to a different dep.

---

### Medium

#### 5. Transitive enrichment is silently skipped in the voice path

`enrich_request_dependency_configs` is only called when `db is not None and user_id is not None`. the voice path passes `user_id=None`, so this enrichment never runs for voice sessions. The transitive chain demo only works via voice, but the `dependency_configs` dict will be empty unless the frontend pre-populates it. The PR description says the demo is "voice-first" — this gap should be explicitly called out, since the transitive expansion feature effectively doesn't work over voice unless the frontend sends all configs upfront.

---

#### 6. `else_branch` rule ordering is not validated (`workflow_rules.py`)

`resolve_workflow_dependencies` evaluates rules in list order and returns on the first match. An `else_branch=True` rule always fires when `matched=False`. If a user places the else rule first in the list, it will always match and subsequent conditional rules are unreachable. The frontend `WorkflowSection.tsx` should enforce else-last ordering, or the backend should validate it. Currently, there's no guard.

---

### Low

#### 7. User-facing workflow prompt strings are duplicated

The `"Before we proceed with..."` / `"Understood. Let's start with..."` strings appear verbatim twice (lines 402–412 and 900–912, 444–446 and 942–946). They should be constants, both to avoid drift and to make them easier to translate.

#### 8. `WorkflowVoiceSyncBridge` always defaults `clear_pending_confirmation=true` (`useWorkflowVoiceSync.ts:39`)

```typescript
clear_pending_confirmation: state.clear_pending_confirmation ?? true,
```

The `true` default means every workflow state publish clears pending confirmations, even when no confirmation is pending. Callers should pass this explicitly — the default should be `false`.

---


## Test Coverage Gaps

- No test for `_resolve_intent_after_clarification_gate` in isolation.
- No orchestration integration test exercising the full C→B→A chain through `process_message` (all existing transitive tests are unit tests on `expand_workflow_dependencies` only).
- If `test_crud_transitive.py` mocks the DB, the full-table-scan issue in `resolve_dependency_agent_config` won't be visible in tests.
- The `else_branch`-first edge case (issue #6) has no test.

---

## Summary

| Severity | # | Items |
|----------|---|-------|
| Critical | 1 | Full table scan in `resolve_dependency_agent_config` |
| High     | 3 | Naive `datetime.now()`, DRY violation in orchestrator, intent source mismatch |
| Medium   | 2 | Voice enrichment silently skipped, else-branch ordering unvalidated |
| Low      | 2 | Duplicated prompt strings, `clear_pending_confirmation` default |

The critical and high items should be addressed before merge. The DRY violation is the one most likely to cause a future regression — the intent mismatch (#4) in particular is a subtle correctness bug that would only manifest when a user breaks an intent lock mid-workflow.

abdulmoeez1225:
# Code Review: PR #157 — Conditional Workflow Engine Round 3


---

## Overview

Implements a conditional workflow engine with static prerequisite steps, `workflow_rules` intent/agent-name branching, and transitive dependency expansion. Also fixes the entry-agent resume bug, voice/UI sync via LiveKit data channel, `Save as new` persistence, and adds the `POST /v1/user-agents` route. The core logic is sound and the architecture maps cleanly to the existing pipeline.

---

## Round 2 Issues — Status After This PR

| # | Issue | Severity | Status |
|---|---|---|---|
| 1 | Full table scan in `resolve_dependency_agent_config` | Critical | **Fixed** — replaced with `_find_agent_config_by_logical_id` using JSON path queries pushed to DB |
| 2 | Naive `datetime.now()` in `update_user_agent` | High | **Fixed** — changed to `datetime.now(UTC)` |
| 3 | DRY violation between `process_message` and `process_message_stream` | High | **Not fixed** — `process_message_stream` received 20+ parallel hunks but both ~500-line methods remain separate; divergence risk persists |
| 4 | Intent source mismatch (`_direct_dependencies_for_request` vs `_resolve_missing_dependencies`) | High | **Fixed** — new code uses `_resolve_missing_dependencies(request, intent)` uniformly in both the clarification gate and main pipeline |
| 5 | Voice enrichment silently skipped (transitive deps never resolved in voice path) | Medium | **Fixed** — worker now resolves `dev_user_id`; `process_message_stream` calls `enrich_request_dependency_configs(request, db, user_id)` |
| 6 | `else_branch` ordering unvalidated in `workflow_rules` | Medium | **Partially fixed** — `_workflow_rules_else_must_be_last` validator added, but only checks position, not count (see Round 3 #6) |
| 7 | Duplicated workflow prompt strings | Low | **Fixed** — extracted to `_workflow_prereq_confirmation_message` and `_workflow_prereq_start_ack` helpers |
| 8 | `clear_pending_confirmation` defaulting to `true` in `useWorkflowVoiceSync.ts` | Low | **Fixed** — default changed to `false`; call sites that need `true` pass it explicitly |

---

## Round 3 Issues — New Bugs Introduced by This PR

### High

#### 1. `onRegister` instability in `WorkflowVoiceSyncBridge` — same class as `useAgentTranscription` bug

**`frontend/src/hooks/useWorkflowVoiceSync.ts`**

`onRegister` is listed in the `useEffect` dependency array. If the caller doesn't wrap it in `useCallback`, the effect re-runs on every render — first calling `onRegister(null)` (tearing down the sync fn) then re-registering it. This creates a race window where workflow Proceed/Cancel clicks silently send nothing to the voice worker. the exact same bug class in `useAgentTranscription` / `handleSegment`. The fix: caller must pass a stable `useCallback` reference, or the effect must omit `onRegister` from its dep array with a comment explaining why.

**Relationship to Round 2:** Independent — `WorkflowVoiceSyncBridge` is an entirely new file introduced in this PR.

---

#### 2. `_workflow_ui_selected_agent` can return empty string

**`backend/src/taskorbit/orchestration/__init__.py`**

```python
def _workflow_ui_selected_agent(request: ConversationRequest) -> str:
    if request.selected_agent:
        return request.selected_agent
    return request.agent_config.id  # ← may be ""
```

`AgentConfig.id` defaults to `""` for unsaved/template agents. When empty, `selected_agent` in `ConversationResponse` becomes `""`, which the frontend treats as no agent selected — causing the UI title to drift back to the intent registry name (e.g. `general_inquiry`).

**Fix:** `return request.agent_config.id or None`

**Relationship to Round 2:** Independent — `_workflow_ui_selected_agent` is a new function introduced in this PR.

---

#### 3. `_DEV_USER_EMAIL` import couples worker process to HTTP auth layer

**`backend/src/taskorbit/worker.py`**

```python
from taskorbit.api.deps import _DEV_USER_EMAIL
```

The worker is a separate process independent of FastAPI. Importing a private name from `api/deps` creates a silent coupling — when JWT auth lands and `deps.py` changes, the worker breaks at import time, not at auth time. The email should live in `config.py` / `Settings` or a shared constants module.

**Relationship to Round 2: ** **Side effect of fixing Round 2 #5** (voice enrichment) — resolving that bug required adding `dev_user_id` lookup in `worker.py`, which introduced this coupling.

---

### Medium

#### 4. Transitive expansion does not traverse `workflow_rules` on intermediate agents

**`backend/src/taskorbit/workflow_rules.py` — `expand_workflow_dependencies`**

```python
# Note: We only expand static dependencies here as branching rules
# are intent-dependent and only evaluated for the active entry agent.
for prereq_id in config.workflow_dependencies:
    visit(prereq_id)
```

If an intermediate agent (B) uses a `workflow_rule` to point at C instead of listing C in `workflow_dependencies`, and the entry agent (A) statically depends on B, the expansion silently omits C. `collect_workflow_dependency_ids` loads all rule deps into `dependency_configs` upfront, but `expand_workflow_dependencies` ignores them during traversal. Any chain where an intermediate agent uses conditional rules is silently incomplete.

**Relationship to Round 2:** Independent — design gap in the new `workflow_rules.py`.

---

#### 5. No uniqueness guard on `POST /v1/user-agents`

**`backend/src/taskorbit/api/routes/user_agents.py` / `crud.py` — `create_user_agent`**

`create_user_agent` does a bare `INSERT` with no uniqueness check on `(user_id, name)`. A user can call "Save as new" multiple times with the same name, creating duplicate rows. The prerequisite dropdown lists agents by name — duplicates appear twice and resolve non-deterministically.

**Fix:** Add a DB-level `UNIQUE(user_id, name)` constraint (surface as 409), or check for an existing name before inserting.

**Relationship to Round 2:** Independent — `POST /v1/user-agents` is a new endpoint.

---

### Low

#### 6. `else_branch` validator allows multiple `else` rules (incomplete fix of Round 2 #6)

**`backend/src/taskorbit/types.py` — `_workflow_rules_else_must_be_last`**

The validator enforces `else_branch` is last but not that there is **at most one**. Two `else: true` rules at the end both pass validation. The first fires and returns, making the second permanently dead code — silently confusing to anyone reading the config.

**Fix:**
```python
if sum(1 for r in rules if r.when.else_branch) > 1:
    raise ValueError("workflow_rules: only one else_branch rule is allowed")
```

**Relationship to Round 2:** **Partial fix of Round 2 #6** — the validator was added to address ordering but the count constraint was missed.

---

#### 7. `_resolve_missing_dependencies` computed twice per turn

**`backend/src/taskorbit/orchestration/__init__.py`**

`_resolve_intent_after_clarification_gate` calls `_resolve_missing_dependencies` internally via `_is_executing_workflow_prerequisite`, then the main pipeline calls it again a few lines later. Both hits are pure computation (no I/O) so correctness is unaffected, but the result from the clarification-gate call should be reused.

**Relationship to Round 2:** Independent — new inefficiency from the clarification gate.

---

#### 8. Worker logs private attributes of `OrchestratorAgent`

**`backend/src/taskorbit/worker.py`**

```python
logger.info(
    "worker_workflow_state_synced",
    selected_agent=agent._current_routed_agent,
    completed_steps=len(agent._completed_workflow_steps),
)
```

`_current_routed_agent` and `_completed_workflow_steps` are private to `OrchestratorAgent`. Reading them directly from `worker.py` breaks encapsulation — renaming either field silently breaks the log line. `sync_workflow_state` should return the values it applied so the worker can log them without reaching into internals.

**Relationship to Round 2:** **Side effect of fixing Round 2 #5** — `sync_workflow_state()` was added as part of voice/UI sync; its logging was written to access private fields directly.

---

## Relationship Map: Round 3 Bugs vs Round 2 Fixes

| Round 3 Bug | Relationship to Round 2 |
|---|---|
| #1 `onRegister` instability | Independent — new file |
| #2 `_workflow_ui_selected_agent` empty string | Independent — new function |
| #3 `_DEV_USER_EMAIL` coupling | **Side effect** of fixing Round 2 #5 (voice enrichment) |
| #4 Transitive expansion ignores `workflow_rules` | Independent — design gap in new code |
| #5 No uniqueness on `POST /v1/user-agents` | Independent — new endpoint |
| #6 Multiple `else` rules pass validation | **Incomplete fix** of Round 2 #6 (else ordering) |
| #7 Double computation | Independent — new inefficiency |
| #8 Worker reads private attributes | **Side effect** of fixing Round 2 #5 (voice enrichment) |

---

## Test Coverage Gaps

- No test for `_resolve_intent_after_clarification_gate` in isolation.
- No orchestration integration test exercising the full C→B→A chain through `process_message` (all transitive tests are unit tests on `expand_workflow_dependencies` only).
- No test for the `else_branch` count > 1 edge case (Round 3 #6).
- No test for `_workflow_ui_selected_agent` when `agent_config.id` is empty.
- No test for duplicate name rejection on `POST /v1/user-agents`.

---

## Overall Summary

| Round | Severity | # | Status |
|---|---|---|---|
| Round 2 | Critical | 1 | Fixed |
| Round 2 | High | 3 | 2 Fixed, 1 Not fixed (DRY) |
| Round 2 | Medium | 2 | 1 Fixed, 1 Partially fixed |
| Round 2 | Low | 2 | Fixed |
| Round 3 | High | 3 | Open |
| Round 3 | Medium | 2 | Open |
| Round 3 | Low | 3 | Open |

**Priority order for this PR:** Round 3 #1 (onRegister race) and #2 (empty selected_agent) must be fixed before merge as both produce observable UI regressions. Round 3 #3 (DEV_USER_EMAIL coupling) should be fixed before the JWT auth migration begins. Round 3 #4 and #5 are functional gaps that will surface during QA demos. The DRY violation from Round 2 remains the highest long-term maintenance risk.

qiblatainf (approved):

Issue #168, opened 2026-06-23
Agent Guadrails need to be added


qiblatainf:
**Describe the bug**
The agent's persona guardrails are not working — the agent gets distracted easily and answers off-topic questions instead of staying within its authorized scope. For example, if someone asks "how to make pizza", the agent starts giving the recipe instead of politely refusing and redirecting back to its intended purpose.

**To Reproduce**
1. Start a conversation with an agent that has persona constraints configured (scope / out-of-scope / refusal template).
2. Ask a clearly off-topic question, e.g. "How do you make pizza?"
3. See the agent answer the off-topic question (gives the full pizza recipe).
4. Notice it never refuses or redirects as it should.

**Expected behavior**
When a request falls outside the agent's authorized scope (or hits a forbidden topic), the agent should politely refuse and redirect using the configured refusal template, instead of answering the off-topic question.

**Screenshots**
<img width="652" height="624" alt="Image" src="https://github.com/user-attachments/assets/74b95463-2033-4788-988c-a238cc10f4d7" />

**Additional context**
The guardrails are currently enforced only through system-prompt text in `with_persona_guardrails()` (`backend/src/taskorbit/integrations/llm/prompts.py`, ticket #69). Since this is prompt-only enforcement, the LLM can ignore it, which is likely the root cause of the drift. Possible directions: strengthen the prompt phrasing/placement, add a dedicated scope/moderation check as a separate step, and add automated tests with off-topic prompts (like the pizza example) to prevent regressions.

Issue #69, opened 2026-05-15
Agent Persona Guardrails


Perimora:
## User story
1. As a system operator
2. I want the agent to stay consistent with its defined persona, even when receiving off-topic or unexpected input
3. So that the agent behaves predictably and aligns with its intended use case.

## Acceptance criteria
* The agent adheres to its configured persona during the entire conversation
* Off-topic or unexpected user input does not cause the agent to switch to an unrelated role (e.g. therapist, assistant, etc.)
* The agent responds to off-topic input in a way that redirects back to its intended domain or politely declines
* Persona constraints are defined as part of the agent configuration (aligned with #6)
* Example scenario (happy path + edge cases) are documented in the team wiki

## Definition of done (DoD)
* General coding standards are met (eg. Formatting, Comments, Logging, no API-Keys in Repo)
* Every PR is peer-reviewed by at least one other developer
* Documentation is updated (eg. System Architecture, SBOM)
* Unit tests for new features have been written

## DoD general criteria
* Feature has been fully implemented
* Feature has been merged into the mainline
* All acceptance criteria were met
* Product owner approved features
* All tests are passing
* Developers agreed to release

PR #190, opened 2026-06-30
fix(guardrails): robust off-topic refusal across voice + new-agent defaults

Asadraza17:
## What & why
#168: agents answered off-topic / forbidden prompts (e.g. a pizza recipe) instead of refusing, most visibly on the voice path, where it could also fire a tool before any refusal. Root cause from #69 was that guardrails were prompt-only and ignorable. This makes the guardrail enforcement robust across all paths and makes new agents safe by construction.

## Changes (three small stages)
1. Voice/stream enforcement: a shared `_scope_refusal` helper runs the code-level scope check before slot extraction and tool dispatch on the stream path (voice goes through this path), so off-topic input is refused before anything executes. Text path already had this.
2. Prompt backstop: when guardrails are present, the system prompt now leads with a top-priority stay-in-role directive (primacy) while keeping the detailed scope/forbidden/refusal block at the end (recency). This is the only guard active where the code check is off (staging/prod).
3. New-agent default: a brand-new "Save as new" user agent with no persona_constraints now gets a sensible default at creation, so it is never stored fully unguarded. Creator-provided constraints are left untouched. Scoped to the user-agent path only; the /v1/agent-configs preset route keeps its exact round-trip contract. Seed templates already carry constraints, so template-cloned agents were already covered.

## Tests
- New stream off-topic refusal test (refusal yielded, no slot extraction, no dispatch, no LLM call) plus an in-scope stream test that still streams the answer.
- In-scope data-extraction still extracts/dispatches with the check ON (no false refusal) on text and stream.
- Prompt-top header ordering test; default-constraints unit tests (inject when absent/empty, preserve partial and full creator constraints).
- Full backend suite green locally.


qiblatainf (commented):
naming convention of the branch is not okay

qiblatainf:
Blocking: 
1. Deterministic guard is OFF in production by default (config.py)
enable_scope_shortcircuit returns True only when app_env == "development". So in staging/prod the code-level scope check is disabled and we fall back to prompt-only enforcement — which is the exact behavior that caused the original "how to make pizza → recipe" bug. As written, this branch doesn't fix the reported issue in prod unless ENABLE_SCOPE_SHORTCIRCUIT=true is set. If this is an intentional safe-rollout default, let's document it and add a follow-up to enable it. Otherwise it should default on.

2. Recipe/food heuristic is brittle + risks false positives (scope_check.py)
The hardcoded food word list (pizza|pasta|burger|...) only generalizes to the demo example and is English-only. It also conflates knowledge ("recipe for pizza") with transactions ("order pizza") — an agent whose scope includes ordering food but lists "recipes" as out-of-scope would wrongly refuse "I want to order a pizza." Suggest narrowing to cooking instructions only (drop buy|order|want|get), or removing the bespoke heuristic in favor of token/regex + prompt.

Non-blocking (discuss)
3. scope (positive scope) is never enforced deterministically — only out_of_scope is. An agent with a tight scope but no out_of_scope gets no code-level guard. Reasonable limitation, but let's document that scope stays prompt-only.

4. default_persona_constraints() sets out_of_scope=["recipes","cooking"] on every new agent — a blank cooking-assistant agent would refuse cooking until overridden. Consider defaulting out_of_scope to empty and relying on the role anchor.

qiblatainf (approved):

PR #198, opened 2026-07-01
feat: add delete conversation history

abdulmoeez1225:
### Delete conversation history
- Adds `DELETE /v1/conversations/{id}` backend route that permanently removes a conversation and all child rows (messages, tool executions, slot extractions) via SQLAlchemy ORM cascade
- Adds `deleteConversation(conversationId)` client function in `conversationApi.ts`
- Restructures `ConversationListItem` card layout: agent name renders on a full-width line with `line-clamp-2` (fixes text truncation), timestamp and a hover-visible trash icon on a second row (fixes overlap)
- Adds a Radix-based `AlertDialog` confirmation before permanent deletion in `HistoryPage` — Cancel keeps the conversation, Delete fires the API call and removes the card from local state without a page reload
- If the deleted conversation was currently selected, the detail panel clears automatically

### Agent configuration delete improvements
- Replaces `window.confirm()` on agent delete with the same `AlertDialog` component for a consistent, styled confirmation flow
- After confirming deletion of the currently loaded agent, automatically switches to the default agent (or the first available one) so the UI is never left in a blank state
- Fixes `Load agent` dropdown becoming unresponsive after a delete: the root cause was a Radix focus-trap conflict between `DropdownMenu` and `AlertDialog` opening simultaneously. Fixed by converting `DropdownMenu` to a controlled component, closing it before opening the dialog (`setLoadMenuOpen(false)`), and deferring the dialog open by one task (`setTimeout(..., 0)`) so Radix fully tears down the dropdown before mounting the new focus trap

### GPU inference fixes (Terraform + docs)
- Fixes Ollama running on CPU instead of GPU: `nvidia-ctk runtime configure --runtime=docker --set-as-default` was inside the `if ! dpkg -l nvidia-container-toolkit` install guard, which is skipped on Deep Learning VM images (toolkit is pre-installed). Moved the configure call **outside** the guard so it always runs, then restarts Docker to apply the daemon config
- Updates `gpu_inference` region/zone defaults in `variables.tf` to `us-east1` / `us-east1-b` to match `terraform.tfvars.example`
- Updates `Documentation/self-hosted-inference.md` with: multi-tenancy environment variables (`OLLAMA_NUM_PARALLEL`, `OLLAMA_MAX_LOADED_MODELS`, `OLLAMA_MAX_QUEUE`, `OLLAMA_FLASH_ATTENTION`), startup model pre-pull, VRAM constraints table, GPU runtime wiring requirement, and a CPU-fallback troubleshooting entry
- Documents zone resource pool exhaustion: `us-east1-b` frequently hits L4 GPU availability limits; fallback to `us-east1-c` or `us-east1-d`

## Test plan

- [x] **Delete conversation — happy path:** open History page, hover a conversation card and click the trash icon, confirm in the dialog → card disappears, detail panel clears, no page reload needed
- [x] **Delete conversation — cancel:** open the dialog, click Cancel → conversation is still in the list
- [x] **Delete selected conversation:** select a conversation to view the transcript, then delete it → detail panel clears and no conversation remains selected
- [x] **Agent delete — AlertDialog:** open Agent Config page, open Load Agent dropdown, click Delete on any non-default agent → styled dialog appears (no browser `confirm()`)
- [x] **Agent delete — currently loaded agent:** delete the agent that is currently active in the header → a different (default) agent loads automatically without needing a page refresh
- [x] **Load agent after delete:** after any agent deletion, open Load Agent dropdown → dropdown opens normally and is fully interactive

### DIFF
--- a/backend/src/taskorbit/api/routes/conversations.py
+++ b/backend/src/taskorbit/api/routes/conversations.py
@@ -6,7 +6,7 @@ import json
 import uuid
 from datetime import UTC, datetime
 
-from fastapi import APIRouter, Depends, HTTPException, Request
+from fastapi import APIRouter, Depends, HTTPException, Request, Response
 from fastapi.responses import StreamingResponse
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
@@ -19,6 +19,7 @@ from taskorbit.database.crud import (
     create_conversation_message,
     create_slot_extractions,
     create_tool_execution,
+    delete_conversation,
     enrich_request_dependency_configs,
     get_conversation,
     get_conversation_history,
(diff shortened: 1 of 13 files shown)


Asadraza17 (approved):
All tested and looks good, approving the PR

PR #200, opened 2026-07-05
Fix/data extraction display

qiblatainf:


backend/src/taskorbit/database/crud.py
Asadraza17: This is the load-bearing path for #183, not `get_slot_extractions`. The History UI
calls GET /v1/conversations/{id}/history (see conversationApi.ts:423), which routes
to `get_conversation_history`, and here we re-query SlotExtraction directly and
return the raw rows on line 335, so the new (tool_id, field_name) dedup added to
`get_slot_extractions` is bypassed on the exact endpoint the History page hits.
For any PROD conversation that already has duplicate slot rows (i.e. the data #183
was filed against), the History screen will still render duplicates after this
merges. Only conversations created post-merge are clean. Two clean fixes: either
apply the same dedup loop inline here before line 363, or have this function reuse
`get_slot_extractions(db, conversation_id)` for the slot query.

backend/src/taskorbit/database/crud.py
```diff
@@ -220,21 +220,44 @@ async def create_slot_extractions(
...
-        extractions = [
-            SlotExtraction(
-                conversation_id=conversation_id,
-                user_id=user_id,
-                tool_id=tool_id,
-                field_name=str(name),
-                field_value=str(value) if value is not None else None,
+        field_names = [str(name) for name in slots]
+        existing_result = await db.execute(
+            select(SlotExtraction).where(
```
Asadraza17, line 259: This is a Python SELECT-then-INSERT/UPDATE with no DB backstop. The SlotExtraction model has no UniqueConstrainton (conversation_id, tool_id, field_name), and the migration doesn't add one. So two
concurrent writers (voice worker in livekit_agent/llm.py:490 racing an SSE /process
in api/routes/conversations.py, or two rapid voice turns on the same call) can both
pass the SELECT before either commits and both db.add(...), silently reintroducing
the exact duplicate rows this PR is fixing. A follow-up ticket adding an Alembic
migration to dedupe legacy + add the UniqueConstraint, plus switching this write to
`pg_insert(...).on_conflict_do_update(...)`, would make the invariant DB-enforced.

backend/tests/test_crud.py
```diff
@@ -237,6 +238,49 @@ async def test_create_slot_extractions_empty_slots_returns_empty(self, db_sessio
         )
         assert rows == []
 
+    @pytest.mark.asyncio
+    async def test_create_slot_extractions_upserts_repeated_fields(self, db_session):
```
Asadraza17, line 242: Both new tests seed via `create_slot_extractions`, which is now an upsert, so they
never produce duplicate (tool_id, field_name) rows and the read-side dedup loop in
`get_slot_extractions` (crud.py:292-296) is never actually exercised. Since that
dedup is the mechanism the docstring says handles legacy PROD data, a future
refactor that drops it would leave all these tests green while legacy conversations
regress. Could we add one test that inserts 2-3 SlotExtraction rows directly via
`db_session.add(...)` for the same (conversation_id, tool_id, field_name) with
staggered `extracted_at`, then asserts get_slot_extractions returns a single row
with the newest field_value?

backend/src/taskorbit/livekit_agent/llm.py
```diff
@@ -482,12 +482,16 @@ def _unanswered_user_turn() -> tuple[str, str] | None:
...
                         content=response.reply.content,
                     )
-                if response.extracted_slots:
-                    tool_id = response.tool_invoked.id if response.tool_invoked else "orchestrator"
+                # Persist slot extractions only when the tool actually fired, and
+                # attribute them to that tool — mirrors the text path. Partial fills
+                # (extracted_slots set but no tool_invoked) are surfaced to the UI for
+                # progress but not saved, which also avoids the generic "orchestrator"
+                # provenance label and duplicate per-turn rows.
+                if response.extracted_slots and response.tool_invoked:
```
Asadraza17, line 502: This voice-path gate change is load-bearing for the fix (without it, voice would
still persist partial fills under a synthetic tool_id and the read-side dedup keyed
on (tool_id, field_name) would treat them as distinct rows), but nothing pins it.
Could we add a small test in test_livekit_agent.py that stubs the orchestrator
response with (extracted_slots={...}, tool_invoked=None) and asserts
create_slot_extractions is NOT called, plus one with both set that asserts it IS
called with tool_id=response.tool_invoked.id? Also worth noting in the PR body that
this silently drops persistence of voice partial fills, so if the call is
interrupted mid-collection the intermediate data is lost, whereas before it was
recoverable under the "orchestrator" provenance.

backend/src/taskorbit/database/crud.py
```diff
@@ -252,14 +275,26 @@ async def create_slot_extractions(
...
+    Deduplicated to one row per ``(tool_id, field_name)`` keeping the latest
+    value, so conversations recorded before the write-side upsert (which stored
+    a new row per turn) still render cleanly in the history view. Fields keep
+    their first-seen order.
+    """
     try:
         result = await db.execute(
             select(SlotExtraction)
             .where(SlotExtraction.conversation_id == conversation_id)
             .order_by(SlotExtraction.extracted_at.asc())
```
Asadraza17: Nit: this ORDER BY has no tie-breaker. `extracted_at` uses `server_default=func.now()`
(transaction-start time), so legacy rows written in one `db.add_all(...)`
transaction share the microsecond timestamp; Postgres doesn't guarantee row order
on ties, so the dict-overwrite winner is nondeterministic and a legacy conversation
could flap between page loads on the /history endpoint (once that path is
deduplicated per Comment 1). One-liner fix:
`.order_by(SlotExtraction.extracted_at.asc(), SlotExtraction.id.asc())`
so the highest-id (last-inserted) row deterministically wins.

Asadraza17 (commented):
Thank you for working on this. The write-side upsert + voice-path
gate change is the right shape for the fix and the code is clean, just a few things needs solving and we are good to go

Asadraza17 (approved):
Round 2 verified. Comments 1, 2, 4, 5 are fully addressed:

/history now delegates to get_slot_extractions (docstring locks in
the contract, double-upsert regression test at test_crud.py:361-381).
UniqueConstraint declared on the model and enforced via Alembic
migration a1b2c3d4e5f6, which collapses legacy duplicates via
DELETE ... NOT IN (SELECT MAX(id) ...) BEFORE creating the constraint,
so ALTER TABLE won't fail on existing prod data.
create_slot_extractions is now a dialect-aware INSERT ON CONFLICT DO
UPDATE (SQLite tests exercise the same on_conflict path as Postgres
prod), so the application-level race window is closed.
Voice-path guard has two focused tests at test_livekit_agent.py:535-579
pinning both branches including tool_id threading.
ORDER BY has an id.asc() tie-breaker with a self-documenting comment.

One residual test-debt note (non-blocking): the read-side dedup loop at
crud.py:294-298 is never exercised against true legacy multi-row shape
because both new tests seed via the upsert and the new UniqueConstraint
makes raw db_session.add() of duplicates raise IntegrityError. Since
the migration already collapses legacy duplicates before adding the
constraint, the loop is largely defense-in-depth at this point. Two
paths for the follow-up: (a) fixture that drops the constraint, seeds
raw duplicates, asserts the collapse; or (b) drop the loop and pin the
migration's dedupe behavior with a migration-level test.

Approving👍

PR #201, opened 2026-07-06
fix: clear user-facing errors on LLM provider failures

Asadraza17:
Closes #197.

## Summary

Before this PR, an OpenAI provider failure (rate limit, auth, generic API error) returned "An unexpected error occurred" to the user because the orchestrator's `except` chain only caught `LLMConfigError` and Python's built-in `TimeoutError`. The intent router additionally swallowed provider failures as low-confidence clarifications, hiding the real cause.

This PR closes the loop end-to-end so every provider failure surfaces a category-specific polite reply on every entry point (POST `/v1/conversations/process`, POST `/v1/conversations/stream` SSE, and the LiveKit voice worker), with the technical detail preserved in the `error` field and a specific Prometheus label.

## What changed

**Backend, provider layer**

- `integrations/llm/openrouter_client.py`: maps `UnauthorizedResponseError`, `TooManyRequestsResponseError`, `ProviderOverloadedResponseError` to the shared `LLMError` taxonomy. New `_bounded_detail` helper caps `exc.body` at 500 chars so oversized upstream bodies cannot bloat logs or the user-visible error.
- `intent/__init__.py`: re-raises `LLMError` instead of masking it as a keyword-matching clarification, so intent-detection failures reach the orchestrator handlers.

**Backend, orchestrator + observability**

- `orchestration/__init__.py`: both text and stream paths catch `LLMConfigError`, `TimeoutError`/`LLMTimeoutError`, `LLMError`, `UnicodeEncodeError`, `ValueError`, `Exception` in that order with distinct metric labels and user messages. An inline note above the chain defends the ordering contract (subclass handlers must precede the `LLMError` base).
- `observability/metrics.py`: pre-registers `llm_provider_error` and `unhandled` in the allow-list so Prometheus exposes them from the first scrape.

**Backend, entry points**

- `api/routes/conversations.py::_sse_generator`: on error events, forwards the polite `reply` alongside the technical `message`, and persists the assistant reply so history stays symmetric with `/process`.
- `livekit_agent/llm.py`: speaks the polite reply through TTS on early-exit AND mid-stream errors (previous behaviour trailed off into silence after partial chunks). Preserves `_locked_intent_name` and `_completed_workflow_steps` on error responses so a transient blip does not strand the session mid-workflow.

**Frontend**

- `lib/conversationApi.ts`: extends the SSE error variant with a nullable `reply` field.
- `components/ConversationalChat.tsx`: renders `event.reply` in the assistant bubble on error, routes it through ElevenLabs TTS for text-chat-with-audio users.

**Tests**

Backend test suite grew across `test_orchestration.py` (stream `LLMTimeoutError`, handler-ordering regression, strengthened mid-stream assertions), `test_openrouter_client.py` (SDK mapping + metric increments + `_bounded_detail` truncation), `test_livekit_agent.py` (voice-worker mid-stream, early-error, workflow-state preservation), `test_stream_route.py` (SSE reply-forwarding + assistant persistence on error), `test_intent_router.py` (`LLMError` re-raise), and an autouse fixture in `conftest.py` that keeps orchestrator LLM JSON calls off the network by default.

**Docs**

New `Documentation/llm-providers.md` covering the error taxonomy, per-provider mapping matrix, orchestrator behaviour, seven Prometheus labels, and how to add a new provider. Linked from root `README.md`.

## Pre-PR review findings addressed

A multi-dimensional pre-PR review (AC compliance, DoD, correctness, security, test coverage, docs quality, regression risk, design, observability, contract surface) plus a completeness critic surfaced the following items. Every one is closed on the branch.

**High**

- SSE forwards the polite `reply` field on error events.
- SSE persists the assistant reply on error so history stays symmetric with `/process`.
- FE `conversationApi.ts` error variant extended with `reply?: string | null`.
- FE `ConversationalChat.tsx` renders the polite reply on error and routes it through TTS.

**Medium**

- Voice worker speaks the polite reply on mid-stream errors, not only on early exit.
- Voice worker preserves `_locked_intent_name` and `_completed_workflow_steps` on error responses.
- Handler-order inline defense comment added above the `except` chain in both text and stream paths.
- Handler-ordering regression test asserts `LLMConfigError` still routes to the `llm_config` bucket.
- Stream `LLMTimeoutError` branch has a dedicated test.
- Mid-stream error test strengthened to assert the metric label, the polite reply, and the `error` field.
- Three LiveKit voice-path tests: mid-stream polite-reply speaking, early-error speaking, workflow-state preservation.
- Two SSE route tests: reply-field forwarding on error, assistant persistence on error.
- Documented orchestrator handler chain expanded from a 4-step summary to the actual 6-step chain.
- Documented Ollama auth mapping corrected to reflect that only 401 is handled today (403 falls through to `LLMAPIError`).

**Low**

- `_bounded_detail` helper caps `exc.body` at 500 characters, with a truncation test.
- Both `llm_timeout` handlers bind the caught exception and log `error=str(exc)`.
- `"unhandled"` metric label pre-registered in the allow-list so Prometheus exposes it from the first scrape.
- OpenRouter metric-increment tests added for `status="auth"` and `status="rate_limit"`.
- OpenRouter body-detail extraction covered for `TooManyRequestsResponseError` and generic `OpenRouterError` (in addition to `ProviderOverloadedResponseError`).
- Documented 8s timeout replaced with the per-provider values (OpenAI 8s, Gemini 10s, Ollama 300s).
- Documented "Adding a new provider" reworded to reflect that `LLMClient` is a `typing.Protocol`, not a subclass base.
- Documented that OpenRouter's `generate_stream` delegates to `generate`, so the SDK-error mapping applies to both entry points on that provider.
- Documented metric-labels table expanded from three rows to all seven `conversation_errors_total` label values.
- Root `README.md` now links to `Documentation/llm-providers.md`, closing the discoverability gap.

## Acceptance criteria

- [x] With OpenAI selected the agent no longer returns "An unexpected error occurred" on provider failures; the user sees a category-specific polite reply on `/process`, on the `/stream` SSE endpoint, and on the LiveKit voice worker.
- [x] Metric label emitted matches the failure category (`llm_timeout` for timeouts, `llm_provider_error` for auth / rate limit / other API errors, `llm_config` for missing config, `invalid_input` for `ValueError`, `unhandled` at the app boundary).

### DIFF
--- a/Documentation/llm-providers.md
+++ b/Documentation/llm-providers.md
@@ -32,7 +32,7 @@ All classes live in `backend/src/taskorbit/integrations/llm/errors.py`.
 |---|---|---|
 | `LLMError` | `Exception` | Base for all provider-layer failures. Catch this to handle "the provider failed for some reason." |
 | `LLMConfigError` | `LLMError` | Static config problem (missing API key, unknown provider, unsupported model). Detected before the request is sent. |
-| `LLMTimeoutError` | `LLMError` | Request exceeded the per-call timeout (default 8s for the chat call). |
+| `LLMTimeoutError` | `LLMError` | Request exceeded the per-call timeout. Values are per-provider: OpenAI 8s (`openai_client._REQUEST_TIMEOUT_SECONDS`), Gemini 10s (`gemini_client._REQUEST_TIMEOUT_MS`), Ollama 300s (`ollama_client._REQUEST_TIMEOUT_SECONDS`). OpenRouter timeouts surface from the SDK's `EdgeNetworkTimeoutResponseError`. |
 | `LLMRateLimitError` | `LLMAPIError` | Provider returned HTTP 429 or an equivalent overload signal. Covers both transient throttling and quota-exhausted-billing on providers that share the 429 code (e.g. OpenAI). |
 | `LLMAuthError` | `LLMAPIError` | Provider returned HTTP 401 or 403 (invalid, revoked, or forbidden key). |
 | `LLMAPIError` | `LLMError` | Any other provider-side error (5xx, malformed responses, empty content, unknown SDK errors). |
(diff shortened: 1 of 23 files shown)


K-Vineesh (approved):
Approved — the core bug is fixed and the implementation is solid. 
One follow-up worth filing: OpenAI errors fall through to the generic 
Exception handler (labeled "unhandled") rather than getting provider-specific 
metric labels like OpenRouter does. The user experience is correct (polite 
reply shown), but the observability story for OpenAI specifically is 
incomplete. Suggest filing a follow-up ticket to map openai.AuthenticationError, 
openai.RateLimitError etc. to the shared LLMError taxonomy the same way 
OpenRouter was done in this PR.

Asadraza17:
Thank you @K-Vineesh for the comment, it has been addressed 👍

PR #202, opened 2026-07-06
fix(#166): resolve provider switching reliability gaps for ElevenLabs/Deepgram

### LINKED ISSUE #166: ElevenLabs and Deepgram provider switching is unreliable
**Describe the bug**  
When the user switches the TTS/STT provider between ElevenLabs and Deepgram, the system state is not reliably updated to reflect the selected provider.

**To Reproduce**  
1. Go to Agent Config.  
2. Switch the TTS or STT provider between ElevenLabs and Deepgram.  
3. Observe that the displayed or persisted provider configuration does not always reflect the selected provider.

**Expected behavior**  
The system state reliably reflects and persists the currently selected TTS/STT provider.

**Additional context**  
Reported by the development team. Not independently validated by PO.

K-Vineesh:
Fixes #166.

## Root cause analysis
The core `_build_tts`/`_build_stt` provider switching logic was correct, 
but three gaps caused the system state to not reliably reflect the selected 
provider:

**1. Stale voice_id persisted on Deepgram switch (PipelineSection.tsx)**
When switching TTS provider from ElevenLabs to Deepgram, `voice_id` was 
not cleared. Deepgram encodes voice in the model name — a leftover 
ElevenLabs `voice_id` sitting alongside `provider: "deepgram"` in the DB 
is inconsistent state. Confirmed from DB inspection: agents with Deepgram 
TTS still had ElevenLabs voice IDs persisted. Fixed by clearing `voice_id` 
to empty string on provider switch.

**2. Text-mode TTS failing silently for Deepgram agents (ConversationalChat.tsx)**
The text-mode read-aloud path forwarded `voiceId` and `modelId` to 
`playSynthesizedSpeech` regardless of provider. The REST `/v1/tts/synthesize` 
endpoint is ElevenLabs-only — passing a Deepgram model name 
(e.g. `aura-2-andromeda-en`) caused ElevenLabs to reject the request and 
text-mode TTS failed silently for any Deepgram-configured agent. Fixed by 
only forwarding these params when `provider === "elevenlabs"`.

**3. Silent fallback with no operator visibility (session.py)**
When `agent_config` was absent or a provider field was missing, `_build_tts` 
and `_build_stt` fell back to env defaults with no log distinguishing 
"using agent config" from "fell back to env default". Added 
`tts_selected`/`stt_selected` log entries in the fallback path with 
`source="env_default"` so operators can confirm which provider is active 
from worker logs.

## Testing
- `npx tsc --noEmit` clean
- Manually verified: switching agent TTS to Deepgram, saving, and starting 
  a voice call — worker logs now show `tts_selected provider="deepgram" 
  source="config"` confirming the correct provider is active


qiblatainf (approved):
1. Add tests for restTtsOptions.
Add two unit tests:
Provider is elevenlabs -> returns { voiceId, modelId } from the config.
Provider is deepgram (or anything non-ElevenLabs) -> returns {}.
That's enough to lock in the behavior. Without it, someone could reintroduce the exact bug you just fixed and CI wouldn't catch it.

Check this out if it's important:
2. Switching Deepgram → ElevenLabs leaves the voice empty.
When you switch to Deepgram you clear voice_id (correct). But when the user switches back to ElevenLabs, voice_id stays "", so the voice dropdown shows blank. It still works (the backend falls back to the env-default voice), but it's a confusing UX. Decide: either populate a default ElevenLabs voice on that switch, or leave it and accept the blank selector. Your call.

K-Vineesh:
Thanks for the review!

1. Added JSDoc to restTtsOptions documenting the elevenlabs-only contract 
   so the behavior is clear to future maintainers.

2. Fixed — switching back from Deepgram → ElevenLabs now restores 
   TTS_VOICE_DEFAULT so the voice dropdown doesn't show blank after a round-trip.

PR #205, opened 2026-07-06
Fix/181 UI cleanup

Asadraza17:
Delivers the #181 UI Cleanup ticket (Asad + Endri), plus a home-screen redesign and a full frontend UI audit pass.

## Original #181 cleanup
- ✅ Removed the dead demo "magic wand" confirmation button
- ✅ Widened the chat layout for wider viewports
- ✅ History cards no longer truncate long agent names
- ✅ Hide the chat-input scrollbar
- ✅ Seeded/default agent voices show readable names instead of raw IDs
- ✅ Call-status pill no longer freezes when the agent/room drops (dead-call recovery with a debounced watchdog)

## Seamless call ending + transitions
- ✅ Ending a call stages a brief "Ending…" wind-down before the summary instead of an instant cut
- ✅ Pre-call, ended, and History surfaces fade/slide in for continuity

## Home dashboard (new)
- ✅ Redesigned the pre-call home screen into a responsive two-column dashboard
- ✅ Left column: featured active agent + Start + a "Switch agent" chip row
- ✅ Right column: live "Recent conversations" panel that deep-links into a specific transcript
- ✅ Right column: "System" status card (Backend / LiveKit)
- ✅ Widened to use the screen; nav now aligns with the cards

## TaskOrbit brand logo
- ✅ Replaced the placeholder React-atom-style mark with the TaskOrbit logo in the nav
- ✅ Added the logo beside the page heading (optimized, bundled asset)

## Frontend UI audit fixes
- ✅ Removed dead code (4 orphaned components + the unused mock-conversations module + stray console.logs)
- ✅ Copy polish (dashboard subheading, input placeholders, "Mic off" status label, History empty state)
- ✅ Fixed an invalid mic-glow color (oklch token wrapped in hsl)
- ✅ Delete control is no longer a button nested inside a button (valid HTML + keyboard focus)
- ✅ Guarded the History deep-link against stale ids
- ✅ Recent panel now has loading/error states and refreshes when a call ends
- ✅ Removed a nested `<main>` landmark

## Layout alignment
- ✅ Aligned the Agent Config page width with Chat and History (max-w-6xl)
- ✅ Fixed the STT/LLM/TTS pipeline overflow (long model names no longer crowd the TTS column)
- ✅ Removed the white halo ring on selected History cards

## Verification
- ✅ `tsc -b && vite build` passes
- ✅ ESLint 0 errors, Prettier clean
- ✅ Live-tested: dashboard, agent switcher, recent deep-links, call-ending, dead-call recovery, History

Closes #181


abdulmoeez1225:
# Code Review — PR #205: Fix/181 UI Cleanup

---

## Overview

A large-scope cleanup PR that delivers the #181 ticket plus a home-screen redesign. The net deletion (−405 lines) is a healthy sign — dead components are removed, debug `console.log` calls cleaned up, and a no-op callback eliminated. The new dashboard layout, dead-call watchdog, and animated transitions are the load-bearing additions.

---

## What's Done Well

**Dead-call recovery (`VoiceSessionBridge.tsx`)**
The debounced watchdog logic is well-designed. The three guard refs (`roomConnectedRef`, `agentJoinedRef`, `connectionLostFiredRef`) correctly prevent false-triggers at startup and ensure the callback fires only once per session. The comment explaining the LiveKit "stays stuck on connecting" behaviour captures the non-obvious invariant correctly.

**Button-in-button fix (`ConversationListItem.tsx`)**
The old `<span role="button">` nested inside a `<button>` was invalid HTML. The new sibling layout with `absolute` positioning and `group-hover` is the correct pattern. The `focus-visible:opacity-100` on the delete button is a good addition — keyboard users can now reach it.

**`ending` state wind-down (`useVoiceCall.ts`)**
Clean approach. `ENDING_BEAT_MS = 600` is a sensible duration, the `endingTimerRef` is properly cleared in all teardown paths (`start`, `restart`, `end`, unmount), and the scale/opacity CSS on the in-call card makes the state visible without being jarring.

**`AgentSwitcher.tsx` — dedup by `agent_id`**
The comment explaining the two-condition active check (`idOf(e) === activeAgentId && e.name === activeAgentName`) is worth keeping — the edge case around renamed copies sharing an `agent_id` is genuinely non-obvious.

**Removal of orphaned components**
`MicButton.tsx`, `RecordingControls.tsx`, `Waveform.tsx`, `TranscriptToolMarker.tsx`, and `mockConversations` are cleanly removed with no apparent loose imports.

---

## Issues to Address

### 1. `<main>` removed without a replacement landmark (`ConversationalChat.tsx`)

```tsx
// Before
<main className="min-h-svh bg-background text-foreground">
// After
<div>
```

The PR description says "Removed a nested `<main>` landmark" but the outer `Layout.tsx` doesn't wrap content in a `<main>` either. Screen readers use `<main>` to skip navigation — without it the entire app loses its primary content landmark.

**Fix:** Either keep `<main>` here, or add it to `Layout.tsx`'s `<Outlet>` wrapper.

---

### 2. `RecentConversations` sorts by string comparison, not date (`RecentConversations.tsx`)

```tsx
// Current — fragile
.sort((a, b) => b.started_at.localeCompare(a.started_at))
```

`localeCompare` on ISO 8601 strings works by accident for the `YYYY-MM-DD` prefix but breaks if `started_at` includes a timezone offset that sorts differently as a string (e.g. `+05:30` vs `Z`).

```tsx
// Fix — use Date comparison
.sort((a, b) => new Date(b.started_at).getTime() - new Date(a.started_at).getTime())
```

---

### 3. `AgentSwitcher` active-chip check inconsistency (`AgentSwitcher.tsx`)

The dedup loop uses both `agent_id` and `name` to find which entry wins (correctly), but the chip's `aria-pressed` and styling only check `idOf(entry) === activeAgentId`:

```tsx
const isActive = idOf(entry) === activeAgentId;  // name not checked
```

A renamed copy and its template can both appear "active" if the dedup collapses them to the right entry but the comparison doesn't use the name. The comment at line 413 explains this ambiguity — the render check should mirror the dedup check:

```tsx
const isActive = isActiveRow(entry);
```

---

### 4. `"ending"` state not in `CallStatus` type

`CallStatusIndicator.tsx` adds a `case "ending"` visual, but `CallStatus` in `types/callState.ts` must include `"ending"` as a valid union member or TypeScript accepts the case silently. Verify the type file is updated; otherwise the switch adds a dead case against an exhaustive union and the `ending` visual never renders.

---

### 5. `loadConversations` may fire redundantly on status flickers

```tsx
useEffect(() => {
  if (call.status === "ended") void loadConversations();
}, [call.status, loadConversations]);
```

If status flickers to `"ended"` more than once (e.g. `ending → ended → idle` path restoring on unmount), this fires a redundant API call. Low severity, but a ref guard (`hasLoadedRef`) or checking `prevStatus` would make it fire-once.

---

## Minor Nits

- **`RecentConversations.tsx`** — "View all {conversations.length} in History" uses the total count. If `conversations.length <= MAX_RECENT`, all items are already shown, making the link redundant. Consider hiding it when `conversations.length <= MAX_RECENT`.
- **`PreCallDiagnostics.tsx`** — `className` prop is now correctly forwarded to `<Card>`. Good fix.
- **Logo `alt=""`** — Correct: decorative image beside a text label should have empty alt.
- **`void loadConversations()`** — using `void` to discard the promise is the right pattern here. No change needed.

---

## Summary

This is a solid cleanup PR with good attention to accessibility (button nesting, `aria-pressed`, `focus-visible`) and a meaningful net line deletion. The three items that **must** be addressed before merge:

1. Restore `<main>` landmark (or add to `Layout.tsx`)
2. Fix `RecentConversations` sort to use `Date` comparison
3. Align `AgentSwitcher` active-chip check with the dedup logic

Asadraza17:
Thanks Abdul. Went through all seven: fixed three, the other four hold up on the current branch.

**Fixed:** date sort (now `Date`-based instead of `localeCompare`), Recent-panel refetch (added a last-write-wins guard so a slow older fetch can't overwrite fresher data), and the "View all" link (persistent underline).

**Keeping as-is:** the `<main>` I removed was nested inside Layout's `<main id="main">`, so there's still exactly one; the active chip dedups by `agent_id` alone so only one can light up; `"ending"` is in the `CallStatus` union and handled everywhere (tsc passes); and the logo `alt=""` is intentional since it sits next to the visible "TaskOrbit" text.

Thanks and sort n refetch were good catches

abdulmoeez1225 (approved):

PR #218, opened 2026-07-10
fix(diagnostics): report LiveKit status from backend config

abdulmoeez1225:
…ontend env var

VITE_LIVEKIT_URL is never set in the production Cloud Run build, so the System card always showed LiveKit as 'not configured' even when the backend's LIVEKIT_URL/API_KEY/API_SECRET were correctly set and voice calls worked (the URL is only ever sent to the browser per-session via POST /v1/livekit/token). /health now reports livekit_configured, and PreCallDiagnostics reads that instead of the unused build-time env var.


shikharthakur2404 (approved):
Manually tested on fix/livekit-not-configured-in-production. /health returns livekit_configured: true. System card shows Backend and LiveKit both ready; technical info shows Configured on backend. Fixes the false "not configured" state from #213.

qiblatainf (approved):

PR #220, opened 2026-07-10
fix(orchestration): dispatch EXTERNAL_API tools before LLM and inject tool_data

Asadraza17:
### Bug 1: no-arg EXTERNAL_API tools never dispatched

`ConversationOrchestrator._run_dispatch_step` only fired a tool when either:
- `no_slots_tool_ready` (END_CALL / AGENT_TRANSFER only), or
- `slots_ready` (needs `intent.required_inputs` populated, only true for DATA_EXTRACTION).

An EXTERNAL_API tool with `args_schema.required == []` matched neither branch and silently skipped dispatch.

### Bug 2: tool_data never reached the LLM

Even when dispatch did run, the text path built the system prompt before dispatch and then called the LLM before dispatch. The stream path also built the system prompt before dispatch. So `tool_data` never made it into the prompt, and the LLM answered from prior knowledge.

## Fix

- New helper `_external_api_requires_args(tool)`.
- Dispatch guard extended with `external_api_no_args_ready`, so a no-arg EXTERNAL_API tool qualifies for immediate dispatch.
- `_build_system_prompt` now accepts an optional `tool_data` and, when present and not aborted, appends the JSON with an instruction telling the LLM to use it instead of prior knowledge.
- `process_message` reordered: dispatch, then build prompt with `tool_data`, then LLM. Same shape as the stream path.
- `process_message_stream` reordered: dispatch, then build prompt with `tool_data`, then stream.

Confirmation flow, refusal short-circuit, and existing DATA_EXTRACTION / AGENT_TRANSFER / END_CALL paths are unchanged; the dispatch return-value plumbing (`llm_text_override`, `early_response`, `response_status`) is preserved.

## Tests

Added 4 tests to `backend/tests/test_orchestration.py`:

- `test_external_api_requires_args_helper` covers True/False branches.
- `test_build_system_prompt_injects_tool_data` asserts the injection + LLM instruction.
- `test_build_system_prompt_omits_tool_data_when_aborted` asserts aborted payloads are dropped.
- `test_no_arg_external_api_dispatches_and_llm_sees_result` is the end-to-end for Carl's tool shape (no required args, GET timeapi.io); asserts dispatch fires and the tool result reaches the system prompt seen by `_call_llm`.

Full backend suite: 713 passed, 0 failed. Ruff clean.


shikharthakur2404 (approved):
Setup: time-test-agent-v2 with a single no-arg Current_Date_and_Time_Tool (GET timeapi.io, args_schema.required: [], extract.dateTime → dateTime).

Voice test: Asked “What time is it in Berlin?” — agent returned live Berlin date/time (e.g. Monday July 13, 12:50 PM), consistent with timeapi.io.

Validates #219 / #220:

No-arg external_api tool dispatches before the LLM
Reply uses live tool data, not training-data guesses
Works on the voice/stream path
Worker logs also show generic_api_request → tool_executed → tool_dispatch_complete for Current_Date_and_Time_Tool. LGTM.


Manual testing completed.

PR #221, opened 2026-07-10
fix: added custom agents to when routed to dropdown

qiblatainf:
## Workflow dropdown fixes (`WorkflowSection.tsx`)

- The "Saved agents" / "Built-in agents" grouping now applies only to the three **dependency** editors — Prerequisite Steps, Allowed Handoffs, Then run prerequisite steps — not "When routed to".
- Styling: trigger stretched full-width (`w-full`), consistent placeholder text across the dependency dropdowns.
- Fixed a dedup bug that silently hid entries: agent lists are built per-row (keyed by DB row id, not `agent_id`), since multiple saved rows can share an `agent_id`.
- Fixed a Radix Select value-collision bug: `SelectItem` values are the unique row id, translated back to `agent_id` on change.
- Fixed a validation state leak: a blocked-add error could get stuck and silently disable Save/Update forever — now cleared on agent switch and when hiding the section that produced it.
- `nameOf()` (the "Selected" badge label) now deterministically prefers a customized copy's name over a template's when both share an `agent_id`, instead of depending on fetch order.
- Legacy rows now use the backend's real `is_customized` flag instead of being hardcoded `true`.

## "Load agent" menu (`AgentConfigPage.tsx`)

- Removed the "My agents" section (redundant with "Saved agents").
- "Saved agents" refreshes immediately after Save/Update.
- **New:** `confirmDelete` now always refreshes the agent list, not just when you delete the agent you currently have open — deleting *any* agent used to leave the list stale.
- **New:** loading a Built-in agent then a Saved agent no longer leaves stale flags behind — "Update" no longer vanishes, and no longer risks silently targeting the wrong agent.
- **New:** duplicate `agent_id` is now blocked live in the Agent ID field, naming the conflicting agent.

## Data-shape fixes (`userAgentsApi.ts`)

- Fixed `loadById`: now reuses `backendToFrontendAgent` instead of assuming the raw backend JSON was already frontend-shaped.
- Made that conversion dual-shape-tolerant (`persona ?? instructions`, etc.) for pre-convention rows.

## Backend (`crud.py` / `types.py` / routes) — new since the original summary

- `/v1/agent-configs` (list, get, create, update, delete) now scoped by `user_id`, matching `/v1/user-agents` — previously any user could see/edit/delete any other user's legacy configs.
- `AgentConfigurationSummary`/`Detail` now expose real `is_customized`/`is_default`.
- New `DuplicateAgentIdError` (409) — backend rejects saves/updates that would give two agents the same `agent_id`, enforced in `create_user_agent`, `update_user_agent`, and `copy_on_write_user_agent`.


abdulmoeez1225:
# Code Review — PR #221: `fix: added custom agents to when routed to dropdown`

## Overview

Reworks agent dropdowns in `WorkflowSection.tsx` to list real saved/built-in agents (grouped) instead of a hardcoded intent list or a deduped-by-`agent_id` map. Also fixes a `loadById` data-shape bug, cleans up the Load Agent dropdown, and makes `backendToFrontendAgent` tolerant of legacy row shapes.

## Findings

### 1. (High) "When routed to" now offers values the backend can never match
The deleted `ROUTED_AGENT_OPTIONS` (`technical_support`, `sales`, `general_inquiry`, `customer_dissatisfaction`) was correct: `workflow_rules.py::_rule_matches()` compares `when.agent_name` against `IntentResult.agent_name`, which is hardcoded in `backend/src/taskorbit/intent/*.py` to exactly those four values — never a custom `agent_id`. The new dropdown (`WorkflowSection.tsx:436-479`) now lists arbitrary saved agents instead, so picking anything but a built-in whose `agent_id` happens to match one of the four strings creates a `workflow_rules` entry that silently never fires.

**Fix:** keep "When routed to" bound to the fixed intent vocabulary (ideally sourced from `AgentRegistry.known_agent_names()`); use the new grouped list only for the two dependency-list editors.

### 2. (Medium) Legacy rows hardcoded as `isCustomized: true`
`WorkflowSection.tsx:320-325` marks every `/v1/agent-configs` row as customized regardless of actual state, which can mislabel default clones as "Saved agents".

### 3. (Medium) `/v1/agent-configs` is unscoped
Removing the `user_id`-scoped "My agents" section in favor of `listAgentConfigs()` leans on `list_agent_configurations()` (`crud.py:533-546`), which has **no `user_id` filter** — returns every row for every user. Invisible today (single dev user, no auth), but a cross-user leak once auth ships.

### 4. (Low) `nameOf()` can show an inconsistent label
`WorkflowSection.tsx:155` picks the first row matching an `agentId`; if a template and a customized copy share an `agentId`, the displayed badge name depends on fetch order.

### 5. (Nit) Unrelated file
`.claude/launch.json` — personal IDE config, doesn't belong in this diff.

## What's solid
- `rowId`/`agentId` split and the Radix Select value-collision fix are correct.
- `loadById` fix (was silently wiping `instructions`) is a real bug fix.
- `backendToFrontendAgent` dual-shape tolerance and the `pendingAddError` reset are good, low-risk fixes.

abdulmoeez1225 (approved):

PR #223, opened 2026-07-10
Feat/203 handoff target dropdown

1Endri:
## Summary
Replaces the free-text agent-transfer target input with a dropdown of available 
agents (built-ins + saved custom agents), so users pick an agent by name instead 
of typing/looking up IDs. Addresses #203.

Closes #203

## What changed
- `AgentTransferEditor` (ToolsSection.tsx) now renders a shadcn Select dropdown 
  listing all agents by name, fetched via `fetchUserAgents()`.
- Selecting an agent stores a stable, backend-resolvable identifier (`entry.id`) — 
  the hyphenated slug for built-ins (e.g. `sales-agent`) and the UUID for custom 
  agents. Chips display the resolved name.
- Already-added agents are shown disabled ("(added)") to prevent duplicates; the 
  list is deduped so a template and its customized copies collapse to one entry.
- Falls back to the original free-text input if the agent fetch fails, so the 
  editor never becomes unusable.
- Fixed a StrictMode/AbortError latch bug that kept the fallback active even when 
  agents loaded successfully.
- Fixed a `targets` desync bug: the editor wrote `tool.targets` but the backend 
  reads `tool.parameters.targets`, and the two could drift. Both are now kept in 
  sync on every add/remove, at load time, and on tool creation.

## AC status
- [x] AC1 — dropdown integrated into the handoff dialogue
- [x] AC2 — lists all accessible agent configurations (built-ins + custom)

## Identifier format (verified)
Stored values match `resolve_transfer_target` (Asad's #212): built-ins resolve by 
slug, custom agents by UUID. Confirmed by tracing the resolver directly.

## Dependency / merge order
⚠️ This depends on #212 (`fix/212-agent-handoff-resolution`), which is **not yet on 
main**. Main still runs the old regex-normalization path. This PR should merge 
**after (or together with) #212**, otherwise main would store the new ID format 
before the resolver that reads it lands. End-to-end handoff verification needs both 
branches together.

## Open items
- [x] Docs update (config-UI section) — in progress.

## Testing done
Manual (verified in browser): dropdown lists each agent once by name, add/remove 
works, duplicates prevented, `parameters.targets` and `targets` stay in sync across 
edits (confirmed via Copy JSON), stored values are the correct resolvable IDs, 
chips persist after save + reload.


Asadraza17 (approved):
Reviewed end to end, this is clean. AC1 (dropdown in the handoff dialog) and AC2 (all agents, built-ins + custom) both met. Nice touches: dedup mirroring AgentSwitcher, storing entry.id as the stable target (resolves directly via resolve_transfer_target), dual-writing targets + parameters.targets so they can't drift, disabling already-added entries, and the free-text fallback when the fetch fails. Docs updated too.

One DoD note, not a blocker: DoD asks for unit tests, but the frontend has no test runner (no vitest, no test script, no FE test CI job), so no FE PR here ships component tests, consistent with how prior FE PRs merged.

Approving!

PR #228, opened 2026-07-11
Stop duplicate tool dispatches and dropped slots across turns

djv03:
## Summary
- **#224**: a tool already confirmed and dispatched could be re-asked for confirmation (and re-dispatched) on a later turn whenever the active tool got re-selected — e.g. after an intent reclassification (#227). `_run_dispatch_step` and the user-initiated end-call shortcut now check a per-conversation "already dispatched" marker (`tool_executions.confirmed`, set precisely at the point of a real dispatch) before asking/dispatching again. A rejected tool is not marked, so it can still be retried.
- **#226**: `SlotExtractor` re-derives the full slot set from the entire conversation history on every turn with no persisted baseline, so a slot filled on an earlier turn could silently regress to missing (or drift to a different value) on a later pass. `_extract_slots` now folds in previously-persisted values for the active tool: a slot once filled is restored if a later pass drops it, and isn't overwritten by a differing value from a later, noisier re-derivation.

Depends on #225 — branched off `fix/voice-confirmation-card` since #224 was explicitly called out there as a known follow-up, and this builds on that confirmation-flow work.

## Test plan
- [x] Backend: `pytest tests/` — full suite passes, 0 regressions
- [x] `ruff check` / `mypy` — no new errors vs. base
- [x] New unit tests for both fixes (dispatch dedup, reject-then-retry, slot restoration/pinning, tool_id scoping, no-db backward compatibility)
- [x] Verified live against the running app: replayed the real intent-flip scenario that previously caused repeated confirmations/duplicate dispatches; confirmed a single `confirmed=true` row per tool across multiple follow-up turns


shikharthakur2404:
Manually tested voice path on fix/224-tool-dispatch-dedup: slots persist across turns (#226), no duplicate confirmation/dispatch after intent shift to order frustration (#224). History: John Alia / John.alia@therategmail.com / 555-1099, 1 tool, single confirmed=true for collect_user_info. LGTM.

shikharthakur2404:
Manually tested voice path on fix/224-tool-dispatch-dedup: slots persist across turns (#226), no duplicate confirmation/dispatch after intent shift to order frustration (#2
Screenshot 2026-07-12 at 01 35 14
24). History: John Alia / John.alia@therategmail.com / 555-1099, 1 tool, single confirmed=true for collect_user_info. LGTM.

shikharthakur2404 (approved):
Manual testing on fix/224-tool-dispatch-dedup (PR228 Test Agent, voice, Ollama qwen3.5:9b)

Passed

#226 — Slot persistence: full_name, email, and phone accumulated across turns without re-asking for name after earlier turns.
#224 — Dispatch dedup: After one approved collect_user_info dispatch, an intent shift ("I'm frustrated with my last order") did not show a second confirmation card. Logs showed tool_already_dispatched_skip for collect_user_info.
Voice confirmation card (BUG-003 / 4aa4137): Amber Approve/Deny card appeared during voice for collect_user_info and end_call with correct action labels. Card appeared once per tool; Approve dismisses the card in voice (confirmation resolved via speech).
Confirmations config (BUG-001 / b1f711e): Agent saved with "confirmations": { "required": true, "tools": [] }; confirmation flow triggered before tool dispatch.
End-call confirmation: end_call confirmation card appeared after "Goodbye, please end the call"; verbal "Yes, go ahead" proceeded to farewell/end flow.
Notes

Voice Approve button dismisses the card only; server-side confirm still requires saying "Yes, go ahead" (by design on voice path).
Ollama latency ~60–90s/turn; STT occasionally garbled email/phone — did not block core #224/#226 behavior.

History may show extra tool_executions rows with confirmed: false (known BUG-002 follow-up; runtime dedup worked).

Verdict: LGTM for merge from manual QA — core #224/#226 fixes and Dhruvin’s voice confirmation + end-call follow-ups behave as intended.

PR #232, opened 2026-07-12
hotfix: refresh cached active agent from backend on load

djv03:
## Summary
- The shared active-agent context (`active-agent-provider.tsx`) caches the whole `AgentConfig` — including `llm.provider`/`llm.model` — in `localStorage`, and only fetched from the backend if that cache was empty. Once any agent was cached, it was served forever (surviving hard reloads), so an edit saved in one tab/session (e.g. switching an agent's LLM provider to Gemini) never reached a session that already had an older snapshot cached.
- Reproduced live: an agent saved in the DB as `gemini/gemini-2.5-flash` was still being sent to the LiveKit worker as `openai/gpt-4o-mini`, which hit that OpenAI key's quota limit and surfaced as a generic "I'm having trouble reaching my language model provider right now" fallback message — looked like an outage, was actually a stale client-side cache.
- Fix: the mount effect now always re-fetches when the cached agent is a real backend-tracked user-agent (`loadedConfigId` starts with `"ua:"`) and overwrites the cached copy with the fresh one, falling back to the cached copy if the backend is unreachable or the agent was deleted elsewhere. No-cache path (first load) behaves exactly as before.

## Test plan
- [x] `tsc --noEmit` — clean
- [x] `eslint` — clean (only a pre-existing, unrelated fast-refresh warning on the same file)
- [x] `npm run build` — succeeds
- [ ] Manual: edit an agent's LLM provider and save in one tab, confirm a second tab/browser with a stale cached copy of that agent picks up the new provider on reload without manually clearing `localStorage`

### DIFF
--- a/frontend/src/components/active-agent-provider.tsx
+++ b/frontend/src/components/active-agent-provider.tsx
@@ -82,20 +82,47 @@ export function ActiveAgentProvider({ children }: { children: ReactNode }) {
   // On mount: load the default user agent from the backend if localStorage
   // has no saved choice. Falls back silently to JOHN_DOE_AGENT if the API
   // is unreachable (e.g. backend not running locally).
+  //
+  // If localStorage DOES have a saved choice that's a real backend-tracked
+  // user-agent (loadedConfigId starts with "ua:"), re-fetch that specific
+  // entry and overwrite the cached copy with it (bug: this provider used to
+  // trust the localStorage snapshot forever once present, so an edit saved
+  // in another tab/browser -- e.g. switching an agent's LLM provider --
+  // never reached a session that already had an older copy cached, even
+  // across hard reloads, and voice/text calls silently kept using the
+  // stale settings). Falls back to the cached copy if the backend is
+  // unreachable or the agent was deleted elsewhere.
   useEffect(() => {
-    if (readFromStorage()) return; // user already has a saved choice
+    const stored = readFromStorage();
     const controller = new AbortController();
-    fetchUserAgents(controller.signal)
-      .then((entries) => {
-        const defaultEntry = entries.find((e) => e.is_default) ?? entries[0];
-        if (defaultEntry) {
-          setAgent(backendToFrontendAgent(defaultEntry));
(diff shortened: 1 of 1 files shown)


Asadraza17 (dismissed):
Requesting changes, the idea is right but the refresh matches the wrong row.

loadedConfigId is built everywhere as ua:${template_id ?? id}, so for any built-in
it's ua:<slug> (default-load line 93, ConversationalChat 811, useAgentHandoff 80).
GET /user-agents returns list_user_agents_merged, which includes BOTH the user's
copy (id=uuid, template_id=slug) AND the pristine template (id=slug). So
find(e => e.id === uaId) with uaId=slug matches the template, not the user's copy.

Failure case, which is the exact bug this PR targets: user switches their agent's
LLM to gemini (saved on their copy), then picks it via the switcher or gets it via
a handoff so loadedConfigId is ua:<slug>. On reload the refresh loads the pristine
template (openai/gpt-4o-mini), setAgent overwrites the gemini config, and the
persist effect writes openai back to storage. So it reverts the customization and
re-introduces the openai quota outage. Works today only because a freshly seeded
copy equals its template; breaks the moment a built-in is customized.

Fix, prefer the user copy and match both id formats:
  const matches = (e) => e.id === uaId || e.template_id === uaId;
  const fresh = entries.find((e) => matches(e) && e.is_customized) ?? entries.find(matches);
  if (fresh) setAgent(backendToFrontendAgent(fresh));

Would flip to Approve once the refresh resolves to the user's copy instead of the
template. Not blocking on a unit test since FE has no runner.

djv03:
Fixed in 7011ac5 — the refresh now prefers the customized copy (`is_customized: true`) matching either `id` or `template_id`, falling back to a plain match only if no customized row exists. Matches the pattern already used in `useAgentHandoff.ts`/`ConversationalChat.tsx` for the same id-vs-template_id ambiguity.

@Asadraza17 ready for another look.

abdulmoeez1225 (approved):

Issue #153, opened 2026-06-17
Voice detection does not terminate


Perimora:
**Describe the bug**
The system does not reliably detect the end of a user's spoken input during voice interaction.

As a result, the system may remain in the listening state instead of forwarding the completed user input to the backend.

**To Reproduce**
* Interact with the system using voice input.
* Speak a user input and then stop speaking.
* Sometimes background noise or the voice interaction workflow prevents the system from recognizing that the user has finished speaking.
* The user input is therefore not forwarded to the backend, or is forwarded only after a significant delay.

**Expected behavior**
The system reliably detects the end of spoken user input and forwards the completed input to the system's backend.

**Actual behavior**
The system sometimes continues waiting for additional speech even though the user has stopped speaking.

**Additional context**
- Shown by dev during review.

PR #234, opened 2026-07-12
diag(voice): log every user_state_changed transition with held_ms

djv03:
## Summary
- Bug 2: live voice calls have shown ~30-40s dead-air stalls where `worker_force_commit_published` (the server-VAD end-of-speech backstop, #153) only fires after a long delay, with nothing but `stt_metrics_collected` heartbeats logged in between. `_on_user_state` previously only logged the `listening` transition that triggers `force_commit`, so a stuck `speaking` state (the suspected cause — background noise parking Silero's speech probability in the activation/deactivation hysteresis band) could only be inferred indirectly from gaps between heartbeat lines.
- Every `user_state_changed` transition is now logged with `old_state`, `new_state`, and `held_ms` (how long the previous state actually lasted, computed from the event's own `created_at` clock), so a stuck state is directly visible in the logs instead of inferred.
- Diagnostics only — no behavior change. This is step 1 of the mitigation plan for Bug 2; deciding between a hard-timeout fallback and further VAD threshold tuning depends on what this logging shows next time the stall reproduces.

## Test plan
- [x] `ruff check` — clean
- [x] New unit test `test_user_state_changed_logs_every_transition_with_held_ms` asserts both the field shape and the `held_ms` computation across two transitions
- [x] Full `test_worker.py` suite passes (25/25)
- [ ] Reviewer: watch `docker compose logs -f worker` for `user_state_changed` lines next time a voice call stalls, to confirm whether VAD is stuck in `speaking` or flapping

### DIFF
--- a/backend/src/taskorbit/worker.py
+++ b/backend/src/taskorbit/worker.py
@@ -177,8 +177,34 @@ async def entrypoint(ctx: JobContext) -> None:
                 audio_duration_ms=round(m.audio_duration * 1000, 1),
             )
 
+    # Timestamp (event-clock, from UserStateChangedEvent.created_at) of the
+    # last user_state_changed transition, so each new transition can log how
+    # long the PREVIOUS state actually held. Diagnostic for bug 2: VAD has
+    # been observed staying in "speaking" for 30-40s straight (no transition
+    # to "listening" at all) before force_commit ever fires, which this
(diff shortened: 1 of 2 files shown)


Asadraza17 (approved):
Looks good, approving!

PR #236, opened 2026-07-12
fix(voice): stop writing a duplicate tool_executions row on dispatch

djv03:
## Summary
- Bug 4: the voice-turn persist block in `llm.py` unconditionally wrote a `tool_executions` row (`confirmed` defaulting to `False`) for every `response.tool_invoked`, even on a turn where the orchestrator's own dispatch path (`_mark_tool_dispatched`) had already written a `confirmed=True` row for that exact same dispatch moments earlier. Every successful tool call ended up with two near-duplicate rows in the audit trail instead of one.
- The write is now skipped specifically when `response.status` is `success`/`ended` — the only two statuses `_mark_tool_dispatched` fires for. Ask (`confirmation_required`) and reject turns still hit this path unchanged, since the orchestrator never marks those as dispatched and this remains the only record of them.
- Cosmetic/audit-trail only — doesn't affect the #224 dedup check (`tool_already_dispatched` already filters on `confirmed=True`), just cuts the redundant row per successful dispatch.

## Test plan
- [x] `ruff check` — clean
- [x] New parametrized tests: `test_llm_node_skips_duplicate_tool_execution_row_on_dispatch` (success/ended → no extra write) and `test_llm_node_persists_tool_execution_row_when_not_yet_dispatched` (confirmation_required/rejected → write still happens)
- [x] Full `test_livekit_agent.py` suite passes (30/30)


shikharthakur2404 (approved):
## Manual test — PASS

Tested on `fix/duplicate-tool-execution-rows` via voice (PR228 Test Agent, `collect_user_info` + confirmations ON).

**Conversation:** `conv-1783936489538-jkll9o`

| Check | Result |
|-------|--------|
| After full flow (confirm + dispatch) | 1× `confirmed=false` for `collect_user_info` |
| Duplicate row on dispatch turn | **None** (was 2× false before fix on PR #228 session) |
| Slots in DB | Alice Johnson, email, phone saved |

Automated tests pass locally. LGTM for Bug 4 / duplicate tool_executions row fix.

PR #237, opened 2026-07-13
feat(#229): smart external_api config form + live Try-it panel

Asadraza17:
Closes #229

## Summary

Follow-up to #219. The external_api tool config form made misconfiguration easy and gave no way to test a tool short of a full conversation. This PR makes the form self-maintaining and adds a live Try-it panel.

## Frontend

- Live detection of {{args.x}} / {{env.X}} refs across URL, query values, header values, and the raw body draft (new pure module lib/externalApiArgs.ts mirroring the backend template regex).
- args_schema auto-syncs from detected refs while machine-shaped; a hand-edited schema is never overwritten. Auto-synced / Edited manually badge, Reset to detected, additive Add missing args.
- Non-blocking mismatch and env-ref warnings; body-on-GET/DELETE warning with one-click Clear.
- Try-it panel: fill the detected args, run the request, see live status, elapsed time, and response body. Nested args are sent as JSON; a stale result clears when the config changes.
- Fixes a pre-existing focus bug in KeyValueEditor (Headers / Query / Response extract / Error mapping): rows were keyed by the editable map key, so typing in a key field remounted the row and dropped focus after one character. Keyed by index now.

## Backend

- New POST /v1/tools/try reuses GenericApiTool.execute, so what you test is exactly what the agent runs.
- SSRF guard (assert_public_url) using `not ip.is_global`: blocks private, loopback, link-local, CGNAT (100.64/10), reserved, multicast, unspecified, and IPv4-mapped IPv6 forms, including the Cloud Run metadata endpoint (169.254.169.254). Wired via an optional url_guard hook so production dispatch stays byte-for-byte unchanged and only Try-it enforces it.
- Try-it strips the config's auth.allowed_env, so an untrusted inline config cannot whitelist and exfiltrate a server env var via {{env.X}}.
- Timeout clamped to 15s, response payload size-capped, and httpx.InvalidURL (not a RequestError) is caught so a malformed URL returns a clean envelope instead of a 500.
- Auth uses the existing dev-user dependency (matches every other endpoint; real JWT is tracked separately).

## Testing

- Backend: 730 passed, 0 failed. The Try-it endpoint has 21 tests (SSRF matrix incl. CGNAT and metadata, env-exfil block, url_guard hook, timeout clamp, URL whitespace strip, InvalidURL-to-envelope, route happy + blocked).
- Frontend: introduces Vitest (27 tests locking the template grammar and schema helpers) plus a CI job. lint / format / build / test all green.
- Manually E2E-verified against a running stack: open-meteo (2 args) and timeapi.io return live data; 169.254.169.254, localhost, and an env-exfil attempt are all blocked.


shikharthakur2404 (approved):
Manual testing complete

Template arg detection + schema auto-sync (lat, lon)

Try-it against open-meteo: HTTP 200, live JSON (~582ms)

SSRF guard blocks 169.254.169.254 and localhost (URL_BLOCKED)

Env ref warning when {{env.MY_KEY}} not whitelisted

Manual schema edit preserved (Edited manually badge)

Header key focus fix verified

Config persists after Update + reload

LGTM

1Endri:
Tests/lint all claims check out:
- Backend: 730 passed / 0 failed / 13 skipped (743 total) confirmed via junit output, matches the PR description exactly
- ruff check / ruff format --check clean
- Frontend type-check / lint clean (had to npm ci locally, lockfile was ahead of my node_modules, unrelated to this PR's correctness)
- Frontend: 27 tests in externalApiArgs.test.ts, all real assertions (whitespace handling, malformed refs, nested-arg collapsing, schema diffing, non-string legacs. 27/27 pass. Note: this covers thetemplate-scan/schema-sync library only; there's no component-level test for the Try-it panel UI itself.
-  SSRF guard (assert_public_url) sound. Testnst the actual ranges rather than reading the code and assuming:
- Private (10.x, 192.168.x, 172.16–31.x), loopback, link-local (incl. 169.254.169.254 metadata), CGNAT (100.64.0.0/10, verified at the exact boundary), IPv4-mapped IPv6 (::ffff:10.0.0.1) all correctly blocked
- Trailing-dot hostnames resolve identically, no bypass
- Redirects aren't followed (httpx.AsyncClient defaults to follow_redirects=False, not overridden), so a 302 can't hopast the check post-validation
- Decimal/hex/octal IP-encoding tricks (2130706433, 0x7f000001) failed to resolve via getaddrinfo in my testing worth a sanity check on the actual Linux prod image since glibc's resolver can be more permissive than what I tested against, but not blocking

One real gap, already called out in the function's own docstring: the guard resolves DNS once to check, then httpx re-resolves independently when it connects a DNS-rebinding TOCTOU window. Partially mitigated for Try-it  (env-allowlist is force-stripped, so a rebinbut a rebind could still make the serverreach an internal service and hand its response back through the Try-it panel. Production dispatch doesn't use this guard at all (admin-configured URLs only, by design), so the exposure is scoped to Try-it. I'd file this as a fast-follow (pin the resolved IP into the actual connection) rather than block the merge on it.

Env-allowlist confirmed both layers: {{env.X}} substitution hard-fails unless X is in allowed_env (no silent skip), and the Try-it endpoint forcibly zeroes allowed_env on every inline config before execution, regardless of what the request body claims so Try-it can't be used to whitelist and exfiltrate a real secret even if the caller tries. Backed by a real test (test_try_endpoint_blocks_env_exfiltration) that round-trips this through the actual HTTP endpoint.

PR #240, opened 2026-07-13
fix: remove agent id issue with dropdowns

qiblatainf:
agent_ids causing clash with dropdown picker


shikharthakur2404 (approved):
LGTM for the dropdown/save/update fixes — this addresses the agent_id clash in the picker and row-level disambiguation looks solid on the frontend.

One follow-up (non-blocking): resolve_dependency_agent_config should strip the row: prefix before get_user_agent, otherwise newly saved workflow_dependencies like row:<  uuid> won’t resolve at runtime. Fine to merge for demo — can be a quick follow-up after.

PR #241, opened 2026-07-13
Feat/tools section add notification

shikharthakur2404:
Summary

Adds clear UI feedback when users add or remove tools in Agent Config — previously a new tool appeared silently with no toast, count, or scroll.
Shows a success toast with the tool type label, a tool count badge on the section header, scroll-to-new-card, and a brief highlight ring on the newly added tool.
Shows a confirmation toast when a tool is removed.
Screenshot 2026-07-13 at 15 20 53

### DIFF
--- a/frontend/src/components/agent-config/ToolsSection.tsx
+++ b/frontend/src/components/agent-config/ToolsSection.tsx
@@ -170,7 +170,8 @@ export function ToolsSection({ value, onChange }: Props) {
               }}
               className={cn(
                 "rounded-lg transition-shadow duration-500",
-                highlightIndex === ti && "ring-2 ring-primary/60 ring-offset-2 ring-offset-background",
+                highlightIndex === ti &&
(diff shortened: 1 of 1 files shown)


1Endri:
Reviewed — checks pass (type-check, lint, 27 vitest green), no secrets, no scope creep, 
notification logic is sound (toasts fire in handlers not render, highlight timer cleans 
up properly).

Two non-blocking notes:
1. Minor edge case: highlightIndex is a raw array index captured at add-time, so if a 
   user removes a different tool within the ~2s highlight window, the ring can land on 
   the wrong card. Low severity — worth a quick fix if easy, not blocking.
2. Branch was a few commits behind main — updating before merge. My #203 dropdown 
   touches AgentTransferEditor in this same file (different region), so I'll confirm a 
   clean merge when #203 lands.

Approving. 👍

1Endri (approved):

PR #246, opened 2026-07-14
fix(orchestration): external_api tools collect caller args + report failures honestly

Asadraza17:
Closes #245 

## Summary

An external_api tool where the caller supplies a value ("tell me the time in Germany") froze on voice, while a tool with a hardcoded value worked.

Root cause: external_api tools never collected their {{args.X}} inputs from the conversation. Slot extraction only fed inputs for data_extraction tools, so external_api args were empty at dispatch and substitution failed with TEMPLATE_INVALID (or, with a populated args_schema, the tool never dispatched). Confirmed live on current main with three cases (query-param arg broke, hardcoded arg worked).

This is separate from #229 (the config-form polish). #229 makes the config correct; this fixes the runtime.

## Changes

- `generic_api.extract_arg_names`: the top-level {{args.X}} names a tool references across url, header values, query values, and body. Ground truth, independent of args_schema.
- `_external_api_requires_args` now keys off those refs, not args_schema.required, so it is correct even when the schema was never filled in.
- Slot-extraction override extended to external_api: the referenced args are collected from the conversation before dispatch, exactly like data_extraction params.
- The tool description is passed to slot extraction as conversion guidance, letting the extractor convert what the user said into the field value (country -> timezone). Scoped so data_extraction extraction stays strictly literal (its prompt is unchanged; a test asserts this).

## Testing

- Backend: 774 passed, 0 failed, ruff clean.
- New tests: extract_arg_names; external_api requires-args / arg-names; an integration test that an external_api arg is extracted and dispatched with the extracted value; a test proving the guidance is absent (and the strict prompt intact) for data_extraction.
- Manually verified end to end against main's code with a real LLM: literal arg (postcode -> town) and conversion arg (Germany -> Europe/Berlin -> correct time), 3/3 consecutive successes on the conversion case.

### DIFF
--- a/backend/src/taskorbit/orchestration/__init__.py
+++ b/backend/src/taskorbit/orchestration/__init__.py
@@ -148,22 +148,32 @@ def _seconds_to_ms(seconds: float | None) -> float | None:
     return round(seconds * 1000, 1) if seconds is not None else None
 
 
-def _external_api_requires_args(tool: ToolDefinition) -> bool:
-    """Return True when an external_api tool's args_schema has required fields.
+def _external_api_arg_names(tool: ToolDefinition) -> list[str]:
+    """Top-level ``{{args.X}}`` names an external_api tool references.
 
-    A "read" tool with no required inputs (e.g. GET current time) can dispatch
-    immediately without any slot-extraction round-trip. A tool that DOES take
-    args (e.g. POST create-user with required fields) still needs its inputs
-    collected, so it goes through the normal slots_ready path.
+    Read from the actual template refs in the request config (URL, headers,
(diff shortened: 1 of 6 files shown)


shikharthakur2404 (approved):
Reviewed and tested locally.

Automated: pytest -k external_api passes; CI checks green.

Manual E2E with Time Agent (external_api + {{args.timezone}}):
- Try it: timezone=Europe/Berlin → HTTP 200, correct dateTime from timeapi.io
- Chat: "Tell me the time in Germany" → agent returned 20:37 Berlin (live API data, not hallucinated)

Fix works as described. LGTM.

